PR #17, opened 2025-09-06
Bug: Zod v3 vs v4 incompatibilities in dashboard 

nick-inkeep:


anubra266:
Already pushed a related PR

PR #31, opened 2025-09-08
Improve CLI

anubra266:


agents-cli/src/utils/tsx-loader.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Dynamically imports a file with TypeScript support
+ * Registers tsx loader for .ts files automatically
+ */
+export async function importWithTypeScriptSupport(filePath: string): Promise<any> {
+  const ext = extname(filePath);
+
+  // For TypeScript files, register tsx loader
+  if (ext === '.ts') {
+    try {
+      const { register } = await import('tsx/esm/api');
```
amikofalvy: @anubra266 does this need to be a dynamic import?

agents-cli/src/config.ts
```diff
@@ -1,365 +1,14 @@
...
-  managementApiUrl?: string;
-  executionApiUrl?: string;
-  outputDirectory?: string;
-  modelSettings?: ModelSettings;
-}
-
-export interface ValidatedConfiguration {
   tenantId: string;
   projectId: string;
-  managementApiUrl: string;
```
amikofalvy, line 22: @anubra266, I think we require both urls? Perhaps this is a bad merge conflict resolution or we are improperly extending types. This type might need to be moved to the core package?

amikofalvy: I see that we still use `executionApiUrl` and `managementApiUrl` elsewhere in this PR. Why is this type different?

anubra266: @amikofalvy it's just this is a different file frombefore, the types didn't change, this file is now for the inkeep.config.ts types, but yeah apiUrl needs to be updated

PR #87, opened 2025-09-10
Create-Agents Package

miles-kt-inkeep:
The create-agents package will create a scaffolded agent framework for local development and testing.

### DIFF
--- a/agents-cli/src/index.ts
+++ b/agents-cli/src/index.ts
@@ -28,10 +28,13 @@ program
 // Create command
 program
   .command('create [directory]')
-  .description('Create a new Inkeep Agent Framework project')
+  .description('Create a new Inkeep Agent Framework Starter Directory')
+  .option('--tenant-id <tenant-id>', 'Tenant ID')
+  .option('--project-id <project-id>', 'Project ID')
   .option('--openai-key <openai-key>', 'OpenAI API key')
   .option('--anthropic-key <anthropic-key>', 'Anthropic API key')
-  .option('--nango-key <nango-key>', 'Nango API key')
+  .option('--manage-api-port <port>', 'Management API port', '3002')
+  .option('--run-api-port <port>', 'Run API port', '3003')
   .action(async (directory, options) => {
     await createCommand(directory, options);
   });
--- a/agents-cli/src/index.ts
+++ b/agents-cli/src/index.ts
@@ -3,6 +3,7 @@ import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { Command } from 'commander';
 import { configGetCommand, configListCommand, configSetCommand } from './commands/config';
+import { createCommand } from './commands/create';
 import { devCommand } from './commands/dev';
 import { initCommand } from './commands/init';
 import { listGraphsCommand } from './commands/list-graphs';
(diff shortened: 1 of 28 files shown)


agents-cli/src/index.ts
```diff
@@ -24,6 +25,20 @@ program
   .description('CLI tool for Inkeep Agent Framework')
   .version(packageJson.version);
 
+// Create command
+program
+  .command('create [directory]')
+  .description('Create a new Inkeep Agent Framework Starter Directory')
+  .option('--tenant-id <tenant-id>', 'Tenant ID')
+  .option('--project-id <project-id>', 'Project ID')
```
amikofalvy, line 33: @miles-kt-inkeep should project id be a parameter here if the directory name is supposed to match id? or are we not enforcing directory name matching the project?

miles-kt-inkeep: I was thinking directory name did not need to match the project because you could have multiple project folders under src/. The cli will prompt for a project id which is used to generate the initial `src/<projectid>/` dir

amikofalvy: Yes. You're right, I was thinking in the wrong scope.

packages/create-agents/README.md
```diff
@@ -0,0 +1,163 @@
+# create-agents
+
+Create an Inkeep Agent Framework project with no build configuration.
```
amikofalvy:
```suggestion
Create an Inkeep Agent Framework scaffold with no build configuration.
```

packages/create-agents/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Quick Start
+
+```bash
+# Interactive mode
+npm init agents
+# or
+npx create-agents
+
+# With project name
```
amikofalvy: I don't want to mix the concepts of repos and Agent Framework object `Project`. Can we make sure these are separate concepts? Can you think of a clear way of doing it?

packages/create-agents/README.md
amikofalvy, line 1: - [ ] Add new package to pnpm workspace
- [ ] Confirm this works with changeset cli
- [ ] turbo.json edits

packages/create-agents/src/index.ts
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+import { program } from 'commander';
+import { createAgents } from '@inkeep/agents-cli/commands/create';
+
+program
+  .name('create-agents')
+  .description('Create an Inkeep Agent Framework project')
+  .version('0.1.0')
+  .argument('[project-name]', 'Name of the project')
```
amikofalvy: consider changing here as well

packages/create-agents/src/index.ts
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+import { program } from 'commander';
+import { createAgents } from '@inkeep/agents-cli/commands/create';
+
+program
+  .name('create-agents')
+  .description('Create an Inkeep Agent Framework project')
```
amikofalvy: and here

packages/create-agents/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "create-agents": "./dist/index.js"
+  },
+  "files": [
+    "dist",
+    "templates"
+  ],
+  "scripts": {
+    "dev": "tsx src/index.ts",
+    "build": "tsc",
+    "test": "vitest",
```
amikofalvy: make sure pnpm test at project root doesn't throw errors because no tests exist for this package

packages/create-agents/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "compilerOptions": {
+      "target": "ES2022",
+      "module": "NodeNext",
+      "moduleResolution": "NodeNext",
```
amikofalvy: elsewhere we use `bundler` for now. @anubra266 is potentially changing this.

amikofalvy (approved):
once feedback is addressed, please ship

PR #88, opened 2025-09-10
Improve error message for timeline if signoz var not set

sarah-inkeep:


sarah-inkeep:
@claude please review

PR #92, opened 2025-09-10
Feature/transfer delegate in description

tim-inkeep:


agents-run-api/src/agents/generateTaskHandler.ts
```diff
@@ -111,6 +112,29 @@ export const createTaskHandler = (
...
+      // Enhance internal relation descriptions with their transfer/delegation info
+      const enhancedInternalRelations = await Promise.all(
+        internalRelations.map(async (relation) => {
+          try {
+            const relatedAgent = await getAgentById(dbClient)({
+              scopes: { tenantId: config.tenantId, projectId: config.projectId },
+              agentId: relation.id,
+            });
+            if (relatedAgent) {
+              const enhancedDescription = await generateDescriptionWithTransfers(
```
amikofalvy: I see this is called in two places? once in genreateTaskHandler and another time when making the agent card. Can we reduce the places it is called to one?

amikofalvy (approved):
reduce repeated usages of the enhancedDescription function.

It's probably an indicator that we need to extract the agent card making functionality into a re-usable function or class that can be called

amikofalvy (approved):

PR #101, opened 2025-09-11
Remove Default Models and Add Interactive Model Configuration

tim-inkeep:
# Remove Default Models and Add Interactive Model Configuration

## Key Changes:
  - Backend: Removed default models from ModelFactory, Agent.ts, and graph-session.ts - now requires explicit project-level configuration
  - CLI: Added interactive provider/model selection to init and push commands when creating projects
  - UI: Enhanced model selectors to show inherited values with "(inherited)" indicator

## Inheritance: structuredOutput/summarizer → base → graph → project (base model required at project level)


robert-inkeep:
@claude

this pr is missing changes to the types??

PR #115, opened 2025-09-11
test fix

shagun-singh-inkeep:


robert-inkeep:
@claude

require doesnt work in dev

PR #215, opened 2025-09-15
Quickstart template support

miles-kt-inkeep:


amikofalvy (dismissed):

packages/create-agents/inkeep-create-agents-0.2.0.tgz
amikofalvy, line 1: @miles-kt-inkeep is this tarfile needed?

PR #254, opened 2025-09-18
Added Openrouter and Vercel Gateway

tim-inkeep:
Reorganizes model provider UX to streamline main provider selection while preserving access to custom models via OpenRouter and Gateway. Changes GPT-5 defaults to GPT-4.1 across the board.

## Changes vs Main
  **Model Provider Organization:**
  - ✅ **Added** enhanced custom model prompts with OpenRouter/Gateway examples (`model-selector.tsx`)
  - ✅ **Added** OpenRouter and Gateway SDK provider support (`ModelFactory.ts`, `package.json`)

  **Default Model Updates:**
  - 🔄 **Changed** OpenAI defaults from `gpt-5-2025-08-07` → `gpt-4.1-2025-04-14`
  - 🔄 **Reordered** model lists to prioritize GPT-4.1 over GPT-5
  - 🔄 **Updated** create-agents quickstart defaults (`create-agents/utils.ts`)

  **Documentation:**
  - ✅ **Added** "Custom Models" section with OpenRouter/Gateway usage examples
  - ✅ **Added** required API key documentation (`OPENROUTER_API_KEY`, `AI_GATEWAY_API_KEY`)

## Result
  ✅ Clean quickstart with 3 main providers (Anthropic, OpenAI, Google)
  ✅ Power users can access any model via `openrouter/model-id` or `gateway/model-id`
  ✅ GPT-4.1 recommended over GPT-5 as OpenAI default


agents-manage-ui/src/components/graph/sidepane/nodes/model-selector.tsx
```diff
@@ -45,7 +45,8 @@ export function ModelSelector({
       const model = models.find((m) => m.value === value);
       if (model) return model;
     }
-    return value ? { value, label: value } : null;
+    // Handle custom models
+    return value ? { value, label: `${value} (custom)` } : null;
```
amikofalvy: @tim-inkeep does this label still make sense given the scope change from custom models to just gateway and open router?

amikofalvy: ditto for the other mentions of custom models below

agents-run-api/src/agents/ModelFactory.ts
```diff
@@ -27,6 +30,11 @@ export class ModelFactory {
         return createOpenAI(config);
       case 'google':
         return createGoogleGenerativeAI(config);
+      case 'openrouter':
+        // Use official OpenRouter provider, cast to handle type mismatch
+        return createOpenRouter(config) as unknown as Provider;
```
amikofalvy: double cast here seems wrong.

`as unknown as Provider`

Is there a better way to clean up these types?

agents-manage-ui/src/components/graph/sidepane/nodes/model-selector.tsx
tim-inkeep: @amikofalvy i think it does because our way of handling custom models is via openrouter and vercel ai gateway. not necessarily supporting custom models. maybe custom is the wrong word? but basically they are our method for supporting other models not in the big 3? what do you think

amikofalvy (dismissed):
When you have addressed all that we discussed on slack, :shipit:

PR #283, opened 2025-09-24
Add Pino Logger

miles-kt-inkeep:


packages/agents-core/src/utils/logger.ts
```diff
@@ -69,7 +184,7 @@ class LoggerFactory {
   /**
    * Get or create a logger instance
    */
-  getLogger(name: string): Logger {
+  getLogger(name: string): PinoLogger {
```
amikofalvy, line 187: @miles-kt-inkeep is it better to have PinoLogger or just Logger? Does it have some special interface that makes it required to have Pino in the name?

miles-kt-inkeep: yeah there are specific pino functions

agents-manage-api/src/logger.ts
```diff
@@ -1,44 +1,2 @@
...
-    return logger.child({ name });
-  }
-  return logger.child({ reqId, name });
-}
-
-export function withRequestContext(reqId: string, fn: Next) {
-  return asyncLocalStorage.run(new Map([['requestId', reqId]]), fn);
-}
+import { getLogger } from '@inkeep/agents-core';
+export { getLogger };
```
amikofalvy, line 2: Let's reexport with service name


```suggestion
export const getLogger = () => getLogger().child({ service: 'agents-manage-api' });
```

amikofalvy (approved):

PR #293, opened 2025-09-24
Add API Key Names and Edits

miles-kt-inkeep:
Adds Api Key names and a key edit form to the ui.


sarah-inkeep (changes requested):
I think there is a bug with the delete dialog, when I click "delete', I get this error `hook.js:608 Error: `DialogPortal` must be used within `Dialog`` so I think we just need to wrap the `DeleteApiKeyConfirmation` with the `Dialog` component.

and small nit but could we style the edit button to look like the one on the project cards (grays instead of blues) and could we move edit above the delete.
Screenshot 2025-09-24 at 2 00 35 PM

PR #410, opened 2025-09-30
highlight clicked on span in activity timeline

shagun-singh-inkeep:


agents-manage-ui/src/components/traces/timeline/timeline-wrapper.tsx
```diff
@@ -139,6 +139,7 @@ export function TimelineWrapper({
   conversationId
 }: TimelineWrapperProps) {
   const [selected, setSelected] = useState<SelectedPanel | null>(null);
+  const [selectedActivityId, setSelectedActivityId] = useState<string | null>(null);
```
sarah-inkeep: instead of a separate state for `selectedActivityId` what do you think about reading from `selected?.item?.id` (just so that there is one source of truth for the selected item)?

sarah-inkeep (changes requested):
looks good! I noticed the gray border here shifts when a span is selected:
Screenshot 2025-09-30 at 1 46 42 PM

and maybe we could make the border color a bit brighter in dark mode:
Screenshot 2025-09-30 at 1 54 12 PM

PR #455, opened 2025-10-02
Separate store actions from state, introduce `useGraphActions` hook

dimaMachina:
While looking at codebase I saw this line

https://github.com/inkeep/agents/blob/94e010a0bc7943e4d052b28972902ef689b0c2be/agents-manage-ui/src/components/graph/sidepane/metadata/metadata-editor.tsx#L57-L67

Unlike using zustand selector at `line 57`, there is `useGraphStore()` without selector, so this component will be always rerenders on all graph store state changes!

This PR adds improvement

1. Forbid


agents-manage-ui/src/features/graph/state/use-graph-store.ts
```diff
@@ -62,7 +63,12 @@ type GraphState = GraphStateData & {
   getEdgeErrors(edgeId: string): GraphErrorSummary['allErrors'];
 };
 
-export const useGraphStore = create<GraphState>()(
+type GraphState = GraphStateData & {
+  actions: GraphActions;
+};
+
+// Not exported, so that no one can subscribe to the entire store
```
dimaMachina:
```suggestion
```

PR #549, opened 2025-10-09
Preserve template literals in replacement strings

miles-kt-inkeep:
When generating placeholders for inkeep pull, template literals will now be preserved. This allows the llm to convert the template literal to contextConfig.toTemplate format.

### DIFF
--- a/.changeset/honest-cars-move.md
+++ b/.changeset/honest-cars-move.md
@@ -1,5 +1,0 @@
----
-"@inkeep/agents-cli": patch
----
-
-template literal handling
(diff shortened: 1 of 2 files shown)


.changeset/honest-cars-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-cli": patch
```
amikofalvy: When you run pnpm changeset are you making sure to select all packages to make sure they are all on the same version?

miles-kt-inkeep: I thought it does this by default?

PR #568, opened 2025-10-12
changed tool execution into tool call and tool result

tim-inkeep:
- Split tool_execution into tool_call and tool_result events for granular start/completion tracking
  - Updated GraphSession, Agent.ts, tests, docs, and CLI to handle new event structure with toolCallId linking


.changeset/perky-cooks-pump.md
```diff
@@ -0,0 +1,6 @@
+---
+"@inkeep/agents-run-api": patch
```
amikofalvy: All packages need a version bump as per our convention.

Remove this file then
`pnpm changeset:quick patch "converted tool execution into tool call and tool result"`

tim-inkeep: question, i run pnpm changeset:quick patch but it only writes patch on one of these is that not what i should do?

PR #596, opened 2025-10-13
baggage fix for conversation id

shagun-singh-inkeep:


robert-inkeep:
@claude don't we need to make changes to the other endpoints too? basically any time we create a conversation id if one isnt passed

robert-inkeep:
@claude okay can you review again

PR #602, opened 2025-10-13
Animate Agent graph based on `ikp-data-operation` event types

dimaMachina:
https://github.com/user-attachments/assets/f06ae03a-ea6e-436e-97fd-bcc736e333a5

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -76,7 +76,7 @@ importers:
         version: link:../packages/agents-core
       '@inkeep/agents-manage-ui':
         specifier: '*'
-        version: 0.15.0(@babel/runtime@7.28.4)(@codemirror/language@6.11.3)(@codemirror/state@6.5.2)(@codemirror/theme-one-dark@6.1.3)(@libsql/client@0.15.15)(@opentelemetry/api@1.9.0)(@opentelemetry/core@2.1.0(@opentelemetry/api@1.9.0))(@types/pg@8.15.5)(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(codemirror@6.0.2)(use-sync-external-store@1.6.0(react@19.1.1))
+        version: 0.15.0(@babel/runtime@7.28.4)(@codemirror/state@6.5.2)(@codemirror/theme-one-dark@6.1.3)(@libsql/client@0.15.15)(@opentelemetry/api@1.9.0)(@opentelemetry/core@2.1.0(@opentelemetry/api@1.9.0))(@types/pg@8.15.5)(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(codemirror@6.0.2)(use-sync-external-store@1.6.0(react@19.1.1))
       '@inkeep/agents-sdk':
         specifier: workspace:^
         version: link:../packages/agents-sdk
@@ -424,8 +424,8 @@ importers:
         specifier: workspace:^
         version: link:../agents-run-api
       '@inkeep/agents-ui':
-        specifier: ^0.14.15
-        version: 0.14.15(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)
+        specifier: ^0.14.17
+        version: 0.14.18(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)
       '@lezer/highlight':
         specifier: ^1.2.1
         version: 1.2.1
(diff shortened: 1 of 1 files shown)


agents-manage-ui/src/components/agent/configuration/edge-types.tsx
```diff
@@ -12,7 +12,7 @@ export enum A2AEdgeType {
 export enum EdgeType {
   A2A = 'a2a',
   A2AExternal = 'a2a-external',
-  Default = 'default',
+  Default = 'custom',
```
dimaMachina, line 17:
```suggestion
  // Built-in 'default' edges ignore the `data` prop.
  // Use a custom edge type instead to access `data` in rendering.
  Default = 'custom',
```

.changeset/spotty-forks-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": minor
+---
+
+Animate Agen graph based on `ikp-data-operation` event types
```
dimaMachina:
```suggestion
Animate Agent graph based on `ikp-data-operation` event types
```

agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -693,14 +693,153 @@ function Flow({
     toolLookup,
   ]);
 
+  useEffect(() => {
+    const onDataOperation: EventListenerOrEventListenerObject = (event) => {
+      // @ts-expect-error -- improve types
+      const { conversationId, timestamp, ...data } = event.detail;
+      console.log('Data operation:', data);
```
sarah-inkeep: can we remove this console log? also I think conversationId and timestamp are unused so maybe we could remove those as well

sarah-inkeep (changes requested):
This looks great, a few notes:
- Could we actually remove the checkmark on the tool, now that I see it in action I think it might be a little confusing (sorry for suggesting we add it)
- For the ball on the edges could we make it so it only travels down the edge once instead of recurring?
- After the tool completes can we show the ball moving back up the edge (in the other direction) to illustrate the data going back to the agent
- There is an `agent_initializing` data-operation, could we add the glow to that agent when that operation happens?
- Could we also add the ball for `transfer` operations?
- Could we make whichever agent or tool is active have the glow effect and then have it stop glowing when it is complete:
for delegation: if agent a delegates to agent b they can both be glowing
but for transfer: if agent a transfers to agent b then only agent b would be glowing

Basically I think we want the ball to reflect directionality of the data and the glow to reflect the active agent or tool if that make sense. Happy to talk through this on a call as well.

agents-manage-ui/src/components/agent/agent.tsx
dimaMachina: > * Could we actually remove the checkmark on the tool, now that I see it in action I think it might be a little confusing (sorry for suggesting we add it)

yep, I think without is better! removed

dimaMachina: > * For the ball on the edges could we make it so it only travels down the edge once instead of recurring?

yep, I think you are right too.
It was tricky to fix it. I need to add effect:

```js
  const ref = useRef<SVGAnimateElement>(null);

  // Without this useEffect, the animation won't start when this component is rendered dynamically.
  useEffect(() => {
    ref.current?.beginElement();
  }, []);
```
to run it once

dimaMachina: > There is an agent_initializing data-operation, could we add the glow to that agent when that operation happens?

hmm, `agent_initializing` sends only global agent id, it doesn't sends `weather-assistant` agentId

Screenshot 2025-10-17 at 14 54 59

dimaMachina: > Could we also add the ball for transfer operations?

Which agent do you use? How can I reproduce this event type? for now I was able to receive only following event types
```
agent_initializing
delegation_sent
tool_call
tool_result
agent_generate
delegation_returned
completion
```

sarah-inkeep: If you click on the edge between the agents and change the relationship to "transfer" instead of delegate, then you should see it.
Screenshot 2025-10-17 at 9 43 24 AM

sarah-inkeep: > hmm, `agent_initializing` sends only global agent id, it doesn't sends `weather-assistant` agentId
ah got it, maybe we can highlight the default agent then, I believe the default agent is the entry point for each graph.

sarah-inkeep (approved):
looks great!

agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -693,14 +693,182 @@ function Flow({
...
```
dimaMachina:
```suggestion
      const data = event.detail;
```

PR #650, opened 2025-10-15
better errors for mcp tools

tim-inkeep:


.changeset/chubby-rings-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+"@inkeep/agents-manage-ui": patch
```
tim-inkeep: i think just use pnpm changeset so we can mark every file gets upgraded so they all have same version number

PR #674, opened 2025-10-15
New models and clean up

robert-inkeep:


tim-inkeep:
changeset, but thats it

PR #689, opened 2025-10-16
Improve full page errors, better styling, improved messaging and pass…

sarah-inkeep:
… errors through instead of catching and overwriting


sarah-inkeep:
@claude please review again

PR #761, opened 2025-10-20
zoom into section of graph where agent is selected

dimaMachina:
https://github.com/user-attachments/assets/c034b0e3-e0bd-4d7d-ad7e-e15bf550cceb

### DIFF
--- a/agents-manage-ui/src/components/agent/agent.tsx
+++ b/agents-manage-ui/src/components/agent/agent.tsx
@@ -295,9 +295,11 @@ function Flow({
   }, []);
 
   // Auto-center agent when sidepane opens/closes
-  // biome-ignore lint/correctness/useExhaustiveDependencies: we want to trigger on isOpen changes
   useEffect(() => {
     // Delay to allow CSS transition to complete (300ms transition + 50ms buffer)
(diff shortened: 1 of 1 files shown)


sarah-inkeep (changes requested):
I noticed that if the sidepane is closed when clicking on a node, it zooms in and then zooms out again.
https://github.com/user-attachments/assets/8f87ae29-9f6d-4e65-9f48-51cd59b735cf

agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -863,13 +865,18 @@ function Flow({
 
   const onNodeClick: ReactFlowProps['onNodeClick'] = useCallback(
     (_, node) => {
-      fitView({
-        maxZoom: 0.9,
-        duration: 300,
-        nodes: [node],
-      });
+      if (isOpen) {
```
sarah-inkeep, line 868: if the user leaves the sidepane open and then clicks on another node, do you think we should zoom into the new node?

dimaMachina: let's see if want to add it for the others in the future

PR #762, opened 2025-10-20
Fix React Flow toolbar buttons overflow caused by nodes and edges

dimaMachina:
## before

https://github.com/user-attachments/assets/e53358b3-db63-4670-af61-23bf95770c35

## after

https://github.com/user-attachments/assets/ead930d9-459a-4189-8991-9746eec30e67

https://github.com/user-attachments/assets/5bff1e15-a42b-4001-8e39-2e691ffbee10

### DIFF
--- a/agents-manage-ui/src/components/agent/toolbar/toolbar.tsx
+++ b/agents-manage-ui/src/components/agent/toolbar/toolbar.tsx
@@ -57,7 +57,13 @@ export function Toolbar({
     <div className="flex gap-2 flex-wrap justify-end content-start">
       {dirty || inPreviewDisabled ? (
         <Tooltip>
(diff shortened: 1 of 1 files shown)


sarah-inkeep (approved):

sarah-inkeep (changes requested):
oops, I just realized the tooltip doesn't show up any more, I think we need to keep the wrapping div on line 55 of `toolbar.tsx`, but otherwise this looks great, thank you!

dimaMachina:
> oops, I just realized the tooltip doesn't show up any more, I think we need to keep the wrapping div on line 55 of `toolbar.tsx`, but otherwise this looks great, thank you!

yep, thank you 🙏 added a comment

PR #782, opened 2025-10-21
Fix frequency fields from disappearing when cleared

sarah-inkeep:


sarah-inkeep:
@claude please re-review

PR #797, opened 2025-10-22
Team Agents

miles-kt-inkeep:
- Any agent within a project can be delegated to from a sub agent within that same project

### DIFF
--- a/.changeset/pretty-gray-possum.md
+++ b/.changeset/pretty-gray-possum.md
@@ -1,0 +1,11 @@
+---
+"@inkeep/agents-cli": minor
+"@inkeep/agents-core": minor
+"@inkeep/agents-manage-api": minor
+"@inkeep/agents-manage-ui": minor
+"@inkeep/agents-run-api": minor
+"@inkeep/agents-sdk": minor
+"@inkeep/create-agents": minor
+---
+
+Team Agents


agents-manage-ui/src/app/[tenantId]/projects/[projectId]/agents/new/page.tsx
```diff
@@ -18,6 +17,7 @@ async function NewAgentPage({
       fetchToolsAction(tenantId, projectId),
       fetchCredentialsAction(tenantId, projectId),
       fetchExternalAgentsAction(tenantId, projectId),
+      fetchTeamAgentsAction(tenantId, projectId),
```
Comment (sarah-inkeep) [inline agents-manage-ui/src/app/[tenantId]/projects/[projectId]/agents/new/page.tsx]: I noticed that we are fetching the team agents here but it doesn't look like the data returned is being used, maybe we can remove if it is not needed?

sarah-inkeep (commented):
just to check is this expected behavior for team agents:

I tried asking for the weather in nyc to an agent that delegates to event planner, it told me it didn't have access to the weather lookup tool:
Screenshot 2025-10-22 at 10 51 45 AM

But if I ask the event planner agent directly, it gives me the answer:
Screenshot 2025-10-22 at 10 51 53 AM

is this just a config issue on my end, like maybe i need to add some prompting to `kind agent`? I added prompting an now it seems to work, should we add something to the default prompt to tell the default sub agent what team agents it has access to?

PR #806, opened 2025-10-22
fix: polling on activities cause chat widget to re-render every second

dimaMachina:


agents-manage-ui/src/hooks/use-chat-activities-polling.ts
```diff
@@ -65,9 +65,9 @@ export const useChatActivitiesPolling = ({
       if (isComponentMountedRef.current && currentConversationId === conversationId) {
         // Only update state if data actually changed (by checking activity count)
         const newCount = data.activities?.length || 0;
-        if (newCount !== lastActivityCount) {
+        if (newCount !== lastActivityCount.current) {
           setChatActivities(data);
-          setLastActivityCount(newCount);
+          lastActivityCount.current = newCount;
```
dimaMachina, line 70: rerendered because `lastActivityCount` was always 0

PR #837, opened 2025-10-28
Schema eval

shagun-singh-inkeep:


robert-inkeep:
@claude can you review this again?

want to make sure its clear the schemas and the relationships are clear

ignore the composite keys and the tenant scoping, ignore any indexing / perf concerns

PR #877, opened 2025-11-03
hierarchy for timeline

shagun-singh-inkeep:


sarah-inkeep (changes requested):
Looks good, a few small nits:

1. Could we add the background color behind the icons so you don't see the gray line underneath, for example:
Screenshot 2025-11-04 at 12 47 09 PM

2. I noticed the vertical line here dips below the horizontal gray line, could we maybe align it so that it ends at the horizontal line and makes a nice corner
Screenshot 2025-11-04 at 12 51 07 PM

3. Maybe we could remove the horizontal line if it's not connected to a vertical line or maybe the vertical line should be extended in this case? I guess I find the floating lines a little confusing / it's hard to tell what their parent is if that makes sense? for example:
Screenshot 2025-11-04 at 12 47 09 PM

4. Maybe when it's collapsed we could remove the vertical line?
Screenshot 2025-11-04 at 1 02 20 PM

5. Could we make the chevron icon slightly larger maybe like 16px by 16px? 
Screenshot 2025-11-04 at 12 51 39 PM

PR #886, opened 2025-11-04
Move initialNodes definition outside Flow component

dimaMachina:
Refactor initialNodes to be a constant outside of Flow component.

### DIFF
--- a/.changeset/young-hats-ring.md
+++ b/.changeset/young-hats-ring.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Move const `initialNodes` definition outside `<Flow />` component


agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -112,6 +112,16 @@ interface AgentProps {
 
 type ReactFlowProps = Required<ComponentProps<typeof ReactFlow>>;
 
+const initialNodes: Node[] = [
+  {
+    id: generateId(),
```
sarah-inkeep, line 117: I think we need to call generateId() from inside of the Flow component otherwise the generated id will be the same for multiple subagents in different graphs which was causing issues, see this thread [https://inkeep.slack.com/archives/C08QXR5CWBH/p1756965511631629](https://inkeep.slack.com/archives/C08QXR5CWBH/p1756965511631629)

dimaMachina: thank you! worth add a comment :)

PR #887, opened 2025-11-05
Revert "run `pnpm format:check` in CI workflow"

robert-inkeep:
Reverts inkeep/agents#884


dimaMachina:
CI check should be fixed in https://github.com/inkeep/agents/pull/888

PR #892, opened 2025-11-05
run `biome check --linter-enabled=false` on CI to check unsorted imports too

dimaMachina:


robert-inkeep:
@claude format:check to me sounds like it shouldn’t ever write

PR #910, opened 2025-11-06
Optimize queries -> only querying for the current page and created new aggregate fast queries for counts

shagun-singh-inkeep:


shagun-singh-inkeep:
@claude review again

PR #934, opened 2025-11-11
dashboard: show dialog if user tries to leave the agent graph page and there are unsaved changes

dimaMachina:
## desktop

image

## mobile

image

## when user closes tab

image

### DIFF
--- a/agents-manage-ui/cypress/e2e/agent.cy.ts
+++ b/agents-manage-ui/cypress/e2e/agent.cy.ts
@@ -13,7 +13,8 @@ describe('Agent', () => {
         const beforeUnloadEvent = new win.Event('beforeunload');
         win.dispatchEvent(beforeUnloadEvent);
       });
-      cy.get('[role=dialog]').contains('Unsaved changes').should('be.visible');
+      cy.get('[role=dialog]').contains('Save changes').click();
+      cy.get('[role=dialog]').should('not.exist');
     });
 
     it('should navigates to the page when changes are discarded', () => {
--- a/agents-manage-ui/cypress/e2e/validation.cy.ts
+++ b/agents-manage-ui/cypress/e2e/validation.cy.ts
@@ -2,10 +2,7 @@
 
 describe('Validation', () => {
   it('for sub agent validate only `prompt` as required field', () => {
-    cy.visit('/');
-    // Click create graph button
-    cy.contains('Create agent').click();
-
+    cy.visit('/default/projects/my-weather-project/agents/new?pane=agent');
     // Wait for app to initialize and click to save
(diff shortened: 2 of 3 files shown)


agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -74,29 +114,6 @@ export type SubAgentExternalAgentConfigLookup = Record<
...
-  nodeTypes,
-  teamAgentNodeTargetHandleId,
-} from './configuration/node-types';
-import { AgentErrorSummary } from './error-display/agent-error-summary';
-import { DefaultMarker } from './markers/default-marker';
-import { SelectedMarker } from './markers/selected-marker';
-import NodeLibrary from './node-library/node-library';
-import { SidePane } from './sidepane/sidepane';
-import { Toolbar } from './toolbar/toolbar';
-
```
dimaMachina, line 99: moved these imports to the top and reformatted

agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -813,6 +838,94 @@ function AgentReactFlowConsumer({
...
+        return;
+      }
+      event.preventDefault();
+      setShowUnsavedDialog(true);
+    };
+    window.addEventListener('beforeunload', handleBeforeUnload);
+    return () => {
+      window.removeEventListener('beforeunload', handleBeforeUnload);
+    };
+  }, [dirty, handleGoBack]);
```
sarah-inkeep: maybe we can extract some of this into a custom hook? I'm a little worried the agent component is getting a bit bloated so would be great if we could extract as much as we can.

agents-manage-ui/src/components/agent/agent.tsx
```diff
@@ -979,6 +1092,35 @@ function AgentReactFlowConsumer({
...
+            </Button>
+            <Button variant="secondary" onClick={proceedWithNavigation} className="max-sm:order-1">
+              Discard
+            </Button>
+            <Button onClick={handleSaveAndLeave} disabled={isSavingPendingNavigation}>
+              {isSavingPendingNavigation ? 'Saving...' : 'Save changes'}
+            </Button>
+          </DialogFooter>
+        </DialogContent>
+      </Dialog>
```
sarah-inkeep: maybe we can make this it's own component also?

sarah-inkeep (changes requested):
I noticed that if I try to close the tab with unsaved changes, then hit 'cancel' on the browser popup, then try to click the 'save' button on our dialog, nothing happens: 
https://github.com/user-attachments/assets/00617bb0-5963-4ab3-9629-04d0d6dfa7f1

PR #965, opened 2025-11-13
added tool history

tim-inkeep:
• Store tool results in conversation history - Tool calls and their results are now persisted in conversation history to provide context for agent-to-agent transfers

  • Add delegation-aware filtering - Conversation history is filtered by delegation context so delegated agents see their own tool results plus top-level conversation history without pollution

  • Update schema and core library - Added tool-result message type support, updated database defaults to include tool results, and filter out _structureHints from stored results

### DIFF
--- a/agents-docs/content/docs/typescript-sdk/memory.mdx
+++ b/agents-docs/content/docs/typescript-sdk/memory.mdx
@@ -1,21 +1,33 @@
 ---
 title: Conversation Memory
 sidebarTitle: Memory
-description: Understand how conversation history is managed and included in the context window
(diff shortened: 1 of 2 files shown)


nick-inkeep:
A follow up item, but consider progressively trimmed tool call results if reaching context limits. Can be a way to compact conversation history without dropping conversation history.

nick-inkeep:
Also, please describe what context delegated agents get vs what context transfer agents get here: https://docs.inkeep.com/typescript-sdk/memory

And how tool call results affect memory.

We should be transparent with users about how memory works.

And lets consider lifting the conversation history token cap? Tool call results can inflate it more quickly.

PR #1006, opened 2025-11-18
Fix console warning: `[Shiki] 10 instances have been created. Shiki is supposed to be used as a singleton, consider refactoring your code to cache your highlighter instance; Or call `highlighter.dispose()` to release unused instances.`

dimaMachina:
see diff with `Hide whitespace`


dimaMachina:
@claude reverify

PR #1031, opened 2025-11-19
Refetch full project after tool result

sarah-inkeep:


sarah-inkeep:
@claude please review

PR #1062, opened 2025-11-20
allow saving JSON schema without `required` field, highlight header of form builder if there are errors, highlight border of JSON Schema editor if there are errors

dimaMachina:
## for builder error state

image

## json schema error state

image


agents-manage-ui/cypress/e2e/components.cy.ts
```diff
@@ -1,14 +1,13 @@
...
 describe('Components', () => {
-  it('should create a new component when adding JSON schema properties with form builder', () => {
+  it('should create a new component when adding JSON schema properties with form builder and without `required` field', () => {
     cy.visit('/default/projects/my-weather-project/components/new');
     cy.get('input[name=name]').type(`test ${Math.random().toString().slice(2)}`);
     cy.get('textarea[name=description]').type('test description');
     cy.contains('Add property').click();
     cy.get('[placeholder="Property name"]').type('foo');
     cy.get('[placeholder="Add description"]').type('bar');
-    cy.get('[role=checkbox').click();
```
dimaMachina, line 11: remove click to `required` checkbox to ensure test pass without `required` array

agents-manage-ui/src/lib/json-schema-validation.ts
```diff
@@ -115,7 +115,7 @@ function validateJsonSchema(schema: any): ValidationError[] {
       });
     }
 
-    if (!schema.required || !Array.isArray(schema.required)) {
+    if (schema.required && !Array.isArray(schema.required)) {
```
dimaMachina, line 118: only show error `Schema must have a "required" array (can be empty)` if user provided `required` as truthy and non array value

PR #1067, opened 2025-11-20
Fix: Data components render without page refresh in Try It mode

amikofalvy:
## Summary

Fixed the issue where data components created during "TRY IT" mode wouldn't render without a page refresh.

### 1. Agent Component ()
- Modified to read , , and  from the agent store instead of just using static props
- Added fallback logic to use props for initial load, but prefer store values when available
- Now when  updates the store (which already happens in the copilot), all components using these lookups get the fresh data

### 2. Playground Component ()
- Added  hook that listens for `ikp-data-operation` events (same pattern as copilot chat)
- When a tool result operation is detected for the current conversation, it refetches data components and updates the store
- Ensures the Playground chat gets the updated component registry immediately after creation

## How It Works

The copilot chat already had a `refreshAgentGraph` mechanism that fetches and updates all lookups in the store. By making the Agent component read from the store instead of just the props, and adding a similar event listener in the Playground, we ensure data components are dynamically updated without requiring a page refresh.

## Result

Data components created during "TRY IT" mode now render immediately without needing to refresh the page! ✨

### DIFF
--- a/agents-manage-ui/src/components/agent/playground/playground.tsx
+++ b/agents-manage-ui/src/components/agent/playground/playground.tsx
@@ -1,11 +1,14 @@
 import { Bug, X } from 'lucide-react';
-import { type Dispatch, useState } from 'react';
+import { type Dispatch, useEffect, useState } from 'react';
 import { toast } from 'sonner';
 import { TimelineWrapper } from '@/components/traces/timeline/timeline-wrapper';
 import { Button } from '@/components/ui/button';
 import { ResizableHandle, ResizablePanel, ResizablePanelGroup } from '@/components/ui/resizable';
+import { useAgentActions } from '@/features/agent/state/use-agent-store';
 import { useChatActivitiesPolling } from '@/hooks/use-chat-activities-polling';
+import { fetchDataComponentsAction } from '@/lib/actions/data-components';
 import type { DataComponent } from '@/lib/api/data-components';
+import { createLookup } from '@/lib/utils';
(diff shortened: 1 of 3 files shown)


agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts
```diff
@@ -1,20 +1,14 @@
...
+ * This maintains backward compatibility while centralizing AI generation logic.
  */
 
-import { ModelFactory } from '@inkeep/agents-core';
-import { streamObject } from 'ai';
 import type { NextRequest } from 'next/server';
-import { z } from 'zod';
-import { fetchDataComponent } from '@/lib/api/data-components';
-import { fetchProject } from '@/lib/api/projects';
+
```
Comment (sarah-inkeep) [inline agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts:9]: maybe we can use the `makeManagementApiRequest` function here? (it already has the logic for api url and add the bypass secret etc)

agents-manage-ui/src/components/agent/playground/playground.tsx
```diff
@@ -47,6 +51,27 @@ export const Playground = ({
     conversationId,
   });
 
+  useEffect(() => {
```
sarah-inkeep, line 54: just to check is this needed? I think the`refreshAgentGraph` function already fetches the full project defs and then updates the store with the new data components? also I think this refetches when the try now chat has a tool_result call but I think we only need to update the data components when the copilot chat has a tool_result call?

PR #1079, opened 2025-11-21
Feat/add-auth-framework

omar-inkeep:
implementing better auth:
using their organization and sso plugins:
https://www.better-auth.com/docs/plugins/organization
https://www.better-auth.com/docs/plugins/sso

adding permissions to all relevant routes (only doing the default 3 roles: owner, admin, and member)

adding a login page

The UI shows errors if there is a permission issue, but there are still no changes to disable / hide components based on user role. This will come later.

The better-auth generated tables are in `packages/agents-core/src/auth/auth-schema.ts`, and are added to the same postgres db, with foreign keys from organization table to projects table, and also from organization table to api_keys table. A lot of the tests were touched due to this new relationship / foreign key, since now we need to have an organization created for each project. The migration file was edited manually to make sure user data stays intact -- I am creating an organization to relate it to each existing project. 

There is an initialization script that runs to create an organization for `process.env.TENANT_ID`, and a user for `process.env.INKEEP_AGENTS_MANAGE_UI_USERNAME` / `env.INKEEP_AGENTS_MANAGE_UI_PASSWORD`

New env vars:
```
BETTER_AUTH_SECRET=<your secret key>
# INKEEP_AGENTS_MANAGE_UI_USERNAME=admin@example.com
# INKEEP_AGENTS_MANAGE_UI_PASSWORD=adminADMIN!@12
DISABLE_AUTH=true
```

`DISABLE_AUTH` is set to true by default. It disables auth, skips the better auth checks, and framework continues to work as before. 

To enable auth/login:  set `DISABLE_AUTH` to false (or comment it out), generate `BETTER_AUTH_SECRET`, and set `INKEEP_AGENTS_MANAGE_UI_USERNAME` / `INKEEP_AGENTS_MANAGE_UI_PASSWORD`

### DIFF
--- a/agents-manage-ui/src/lib/api/api-config.ts
+++ b/agents-manage-ui/src/lib/api/api-config.ts
@@ -32,7 +32,7 @@ async function makeApiRequestInternal<T>(
   options: RequestInit = {}
 ): Promise<T> {
   const url = `${baseUrl}/${endpoint}`;
-  
+
   let cookieHeader: string | undefined;
   if (typeof window === 'undefined') {
     try {
@@ -41,14 +41,12 @@ async function makeApiRequestInternal<T>(
       const allCookies = cookieStore.getAll();
       // Only forward Better Auth cookies for security
       const authCookies = allCookies.filter((c) => c.name.startsWith('better-auth.'));
-      cookieHeader = authCookies
-        .map((c) => `${c.name}=${c.value}`)
-        .join('; ');
+      cookieHeader = authCookies.map((c) => `${c.name}=${c.value}`).join('; ');
     } catch {
       // Not in a server component context, skip cookie forwarding
     }
   }
-  
+
   const defaultHeaders: HeadersInit = {
     'Content-Type': 'application/json',
     ...options.headers,
@@ -67,13 +65,13 @@ async function makeApiRequestInternal<T>(
     if (!response.ok) {
       const errorData = await response.json().catch(() => ({
         code: 'unknown',
-        detail: 'Unknown error occurred',
+        message: 'Unknown error occurred',
       }));
 
       throw new ApiError(
-        {
-          code: errorData.code || 'unknown',
-          message: errorData.detail || errorData.message || 'Unknown error occurred',
+        errorData.error || {
+          code: 'unknown',
+          message: 'Unknown error occurred',
         },
         response.status
       );


agents-manage-api/src/routes/agent.ts
```diff
@@ -25,8 +24,27 @@ import {
...
+const app = new OpenAPIHono<{ Variables: BaseAppVariables }>();
+
+app.use('/', async (c, next) => {
+  if (c.req.method === 'POST') {
+    return requirePermission({ agent: ['create'] })(c, next);
+  }
+  return next();
+});
+
+app.use('/:id', async (c, next) => {
```
amikofalvy, line 39: I have longterm concerns over defining API Authorization so far away from the endpoint.

I think this creates a false sense of security for devs working on routes. Devs will expect that the API Authorization is already defined for their route or that the Authorization is handled by some higher level middleware and forget to attach Authorization to their endpoints leaving their endpoints accidentally unsecured.

I would prefer defining the authorization directly on each endpoint or make all routes inaccessible/unauthorized by default unless explicitly claims/scopes are provided.

 Additionally, it would be best to develop a method by which adding the Authorization to each endpoint will enhance the generated OpenAPI spec with the required scopes. https://swagger.io/docs/specification/v3_0/authentication/

Additionally, we need to make sure that validated claims are passed into the Data Access Layer which was built to do an additional layer of Authorization/validation.

agents-manage-api/src/middleware/require-permission.ts
```diff
@@ -0,0 +1,72 @@
...
+      await next();
+    } catch (error) {
+      if (error instanceof HTTPException) {
+        throw error;
+      }
+
+      throw createApiError({
+        code: 'internal_server_error',
+        message: 'Failed to verify permissions',
+      });
```
amikofalvy, line 70: Is this necessary? I would have thought that our generic uncaught error response would make sure that a 500 response is returned correctly.

agents-manage-api/src/middleware/session-auth.ts
```diff
@@ -0,0 +1,28 @@
+import { createMiddleware } from 'hono/factory';
+import { HTTPException } from 'hono/http-exception';
+
+export const sessionAuth = () =>
+  createMiddleware(async (c, next) => {
+    try {
+      const user = c.get('user');
+
+      if (!user) {
+        throw new HTTPException(401, {
```
amikofalvy, line 10: should we use the createApiError function in this file as well?

agents-manage-api/src/middleware/tenant-access.ts
```diff
@@ -0,0 +1,55 @@
...
+          message: 'Access denied to this organization',
+        });
+      }
+
+      c.set('tenantId', tenantId);
+      c.set('tenantRole', organizationAccess.role);
+
+      await next();
+    } catch (error) {
+      if (error instanceof HTTPException) {
```
amikofalvy, line 46: can we get rid of this catch too?

PR #1086, opened 2025-11-21
fix: show proper 403 access denied error messages instead of generic internal server error

amikofalvy:
## Summary

Fixes the issue where users accessing tenants they don't have permission for would see generic "internal server error" messages instead of proper "Access denied" errors.

### Backend (agents-manage-api)
- Updated `tenant-access.ts` middleware to use `createApiError()` for proper RFC 7807-compliant error responses
- Updated `session-auth.ts` middleware to use `createApiError()` for consistent error formatting
- Errors now return proper JSON with `code`, `status`, and `message` fields

### Frontend (agents-manage-ui)
- Enhanced `FullPageError` component to properly detect and display `ApiError` status codes
- Added `isApiError()` type guard to detect `ApiError` instances
- Fixed error serialization in Next.js server components to preserve status codes across server/client boundary
- Removed hydration-causing offline check

## Testing

- ✅ Verified user with access to `tenant2` can access it successfully
- ✅ Verified same user cannot access `default` tenant and sees proper "Access denied to this organization" message
- ✅ Error messages now correctly display authorization issues instead of generic errors
Screenshot 2025-11-20 at 10 25 32 PM

## Changeset

Created patch-level changeset for this bug fix.

### DIFF
--- a/agents-manage-api/src/middleware/require-permission.ts
+++ b/agents-manage-api/src/middleware/require-permission.ts
@@ -42,9 +42,17 @@ export const requirePermission = <Env extends MinimalAuthVariables = MinimalAuth
     try {
       // Type assertion needed due to better-auth's complex type inference
       // hasPermission is provided by the organization plugin
-      const result = await auth.api.hasPermission({
+      // https://www.better-auth.com/docs/plugins/organization#access-control-usage
+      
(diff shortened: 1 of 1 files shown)


amikofalvy:
Merged in another PR

PR #1131, opened 2025-11-24
Automatically update API reference for `manage-api` and API reference for `run-api`

dimaMachina:
- [x] auto update API reference for `manage-api`
- [x] **NEW** API reference for `run-api`
  image

- [x] updates all fumadocs dependencies
- [x] refactor all scripts from `js` to `ts`, thanks to  `--experimental-strip-types` flag
- [x] polish README 
- [x] refactor `tailwind.config.mjs` to `ts`
- [x] add `"type": "module"` in `package.json` to avoid specifying `.mjs` extension (like `postcss.config.mjs`)
- [x] `pnpm generate-openapi` and `pnpm validate-link` in `prebuild` script
- [ ] fix a lot of broken urls (should be done in another PR)
  image

PR #1147, opened 2025-11-25
docs: do not hide 'on this page' text in table of contents

dimaMachina:
I don't think we should hide this toc header

image


dimaMachina:
done in https://github.com/inkeep/agents/pull/1131

PR #1150, opened 2025-11-25
fix `TypeError: Cannot destructure property 'relationshipId' of 'data.details' as it is undefined.`

dimaMachina:
it seems that details can be `undefined` for `error` state

### DIFF
--- a/.changeset/modern-bees-notice.md
+++ b/.changeset/modern-bees-notice.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+fix `TypeError: Cannot destructure property 'relationshipId' of 'data.details' as it is undefined.`


agents-manage-ui/src/features/agent/state/use-agent-store.ts
```diff
@@ -446,7 +446,7 @@ const agentState: StateCreator<AgentState> = (set, get) => ({
             };
           }
           case 'error': {
```
speciankarel2-code, line 448:
```

```

speciankarel2-code, line 448:
```suggestion
           case 'error': {[]()_**@[]()@amikofalvy **_
```

speciankarel2-code (approved):

PR #1158, opened 2025-11-25
feat(manage-ui): add /logout URL endpoint for debugging

amikofalvy:
## Summary
- Adds a GET `/logout` route to manage-ui that can be visited directly in the browser to log out
- Useful for debugging login/logout issues without needing the UI to load or JavaScript to execute
- Clears better-auth session cookies both server-side and client-side

## How to use
- Visit `/logout` to log out and redirect to `/login`
- Visit `/logout?redirect=/some-path` to log out with a custom redirect

## Test plan
- [ ] Visit `/logout` while logged in - should clear session and redirect to `/login`
- [ ] Visit `/logout` while logged out - should redirect to `/login` without errors
- [ ] Visit `/logout?redirect=/` - should clear session and redirect to root
- [ ] Verify cookies are cleared in browser dev tools after logout

### DIFF
--- a/agents-manage-ui/src/app/logout/route.ts
+++ b/agents-manage-ui/src/app/logout/route.ts
@@ -79,16 +79,47 @@ export async function GET(request: NextRequest) {
 
   // Clear all known better-auth cookies by setting them to expire immediately
   // This ensures cookies are cleared even if the server-side sign-out fails
+  // We must match the attributes used when setting the cookies (Secure, SameSite, etc.)
   const cookiesToClear = new Set([
     ...BETTER_AUTH_COOKIES,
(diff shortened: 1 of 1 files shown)


agents-manage-ui/src/app/logout/route.ts
```diff
@@ -0,0 +1,69 @@
...
+  'better-auth.two_factor',
+];
+
+/**
+ * GET /logout
+ *
+ * Debug endpoint to log out by visiting a URL directly.
+ * Clears better-auth session cookies and redirects to login.
+ */
+export async function GET(request: Request) {
```
dimaMachina: maybe it should be just server-action? wdyt?

PR #1174, opened 2025-11-26
dashboard: add spinner in `Save changes` button in agent workflow / remove `inPreviewDisabled` prop from `<Toolbar />` component

dimaMachina:
`agents/new` page was removed, we no longer need `inPreviewDisabled` prop


agents-manage-ui/src/components/ui/spinner.tsx
```diff
@@ -0,0 +1,16 @@
+import { Loader2Icon } from "lucide-react"
```
dimaMachina, line 1:
```
npx shadcn@latest add spinner
```

PR #1176, opened 2025-11-26
dashboard: move `Ctrl+S` event listener from `<Toolbar />` component to `useAgentShortcuts()` hook

dimaMachina:
I think it make sense to move this listener, especially since we already have dedicated hook

@sarah-inkeep Also I noticed, `undo`/`redo` shortcuts doesn't work when I remove node/edges, should I create a ticket?

### DIFF
--- a/agents-manage-ui/cypress/support/e2e.ts
+++ b/agents-manage-ui/cypress/support/e2e.ts
@@ -20,7 +20,8 @@ Cypress.on('uncaught:exception', (err) => {
   if (
     // Promise from monaco-editor
     err.message.includes('  > Canceled') ||
-    err.message.includes('  > ResizeObserver loop completed with undelivered notifications.')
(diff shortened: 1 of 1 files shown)


agents-manage-ui/src/features/agent/ui/use-agent-shortcuts.ts
```diff
@@ -1,41 +1,45 @@
...
       }
       if (e.key === 'Delete' || e.key === 'Backspace') {
-        // Let inputs handle backspace/delete
-        const target = e.target as HTMLElement | null;
-        const isEditable =
-          target &&
-          (target.tagName === 'INPUT' ||
-            target.tagName === 'TEXTAREA' ||
-            (target as any).isContentEditable);
-        if (!isEditable) {
```
dimaMachina, line 28: we don't need this since we check keys only for `react-flow__node` class which is never input, textarea or `isContentEditable`

PR #1177, opened 2025-11-26
updated relation schema for delegations

tim-inkeep:


amikofalvy (commented):
It's possible we can get unexpected behavior by sending the default message "Delegated task"

PR #1182, opened 2025-12-01
use original fumadocs styling for docs

dimaMachina:


agents-docs/src/app/global.css
```diff
@@ -91,111 +91,107 @@
...
-  word-break: inherit !important;
-}
-
-.type-table {
-  @apply gap-1;
-}
-.type-table div[data-state] button {
-  @apply flex gap-2;
-}
+/*body {*/
```
dimaMachina, line 103: I've commented all custom css overrides for now

agents-docs/src/app/[[...slug]]/page.tsx
```diff
@@ -31,34 +26,17 @@ export default async function Page(props: PageProps<'/[[...slug]]'>) {
...
       </div>
       {page.data.description && (
-        <div>
-          <Markdown
-            text={page.data.description}
-            components={{
-              p: (props) => <p {...props} className="text-lg text-fd-muted-foreground" />,
-            }}
-          />
-        </div>
```
Comment (dimaMachina) [inline agents-docs/src/app/[[...slug]]/page.tsx:61]: frontmatter description should **never** contain markdown, because it's injected in Next.js metadata description, so this compilation is unneeded

PR #1192, opened 2025-12-01
feat: add context tracking evaluation system for trace viewer

amikofalvy:
## Summary

- Add a context tracking evaluation system that shows token usage breakdown for each component of the LLM context window in the trace viewer
- Uses character-based token approximation (~4 chars/token) which works universally across all models (OpenAI, Anthropic, Gemini, custom)
- Records context breakdown as OTEL span attributes on `agent.generate` spans

### Backend (agents-run-api)
- **token-estimator.ts**: New utility for character-based token estimation with `ContextBreakdown` interface
- **Phase1Config.ts / Phase2Config.ts**: Return `AssembleResult` with prompt and breakdown from assembly
- **SystemPromptBuilder.ts**: Pass through breakdown data
- **Agent.ts**: Record 12 context breakdown attributes as OTEL span attributes

### Shared (agents-core)
- **otel-attributes.ts**: Add `context.breakdown.*` attribute keys

### Frontend (agents-manage-ui)
- **context-breakdown.tsx**: New visualization component with stacked bar chart and detailed list
- **conversation-detail.tsx**: Integrate breakdown component into summary cards
- **route.ts**: Parse breakdown attributes from span data

## Components Tracked

| Component | Description |
|-----------|-------------|
| System Prompt Template | Base XML template tokens |
| Core Instructions | Core prompt instructions |
| Agent Prompt | Agent-level context |
| Tools Section | MCP, function, and relation tools |
| Artifacts Section | Artifact definitions |
| Data Components | Phase 2 data component instructions |
| Artifact Components | Artifact creation instructions |
| Transfer Instructions | Agent transfer guidance |
| Delegation Instructions | Agent delegation guidance |
| Thinking Preparation | Extended thinking instructions |
| Conversation History | Chat history tokens |

## Test plan

- [ ] Verify type checking passes across all packages
- [ ] Test trace viewer shows context breakdown card for agent generation spans
- [ ] Verify breakdown values are reasonable (sum to total)
- [ ] Test with different agent configurations to ensure all components are tracked

### DIFF
--- a/.changeset/hot-rice-matter.md
+++ b/.changeset/hot-rice-matter.md
@@ -1,5 +1,0 @@
----
-"@inkeep/agents-manage-ui": patch
----
-
-Update tool parsing to match tool name changes in updated mcp


shagun-singh-inkeep:
@claude review

agents-run-api/src/utils/token-estimator.ts
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * Uses character-based approximation (~4 characters per token) which:
+ * - Works universally for all models (OpenAI, Anthropic, Gemini, custom)
+ * - Requires no external dependencies
+ * - Is fast (simple string length calculation)
+ * - Is accurate enough for relative comparisons between context components
+ */
+
+// Average ~4 characters per token across most LLM tokenizers (BPE-based)
+const CHARS_PER_TOKEN = 4;
```
amikofalvy, line 12: @shagun-singh-inkeep @robert-inkeep, @tim-inkeep  Do you think this simple token estimator will suffice for now? If we can find a better token estimator later we can replace this module with the third party options or options that are specific to individual models.

agents-run-api/src/utils/token-estimator.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Breakdown of estimated token usage for each context component.
+ * All values are approximate token counts.
+ */
+export interface ContextBreakdown {
```
amikofalvy, line 18: My initial design had these explicit items listed in the breakdown; however, now I have concerns over the long term feasibility of this approach.

This explicit breakdown creates a second place where we are defining components.

Previously, we have defined SystemPrompt Versions, we are using V1 (https://github.com/inkeep/agents/tree/main/agents-run-api/templates/v1). I think these context breakdowns should be defined/derived by the version file in a single location.

This will make maintaining the breakdowns easier.

tim-inkeep: meaning that we could have different breakdowns  with different components depending on the version of the system prompt?

PR #1217, opened 2025-12-05
fix(cypress): use correct event types for sidebar hover test

amikofalvy:
## Summary

- Fixed flaky Cypress test `should temporarily expands on hover and collapses again on blur` in `sidebar.cy.ts`
- Changed `mouseover` to `mouseenter` and `mouseout` to `mouseleave` to match the sidebar handler's expected event types

## Root Cause

The sidebar's `handleHover` callback checks `event.type === 'mouseleave'` to determine if the mouse left the sidebar:

```typescript
const isBlur = event.type === 'mouseleave';
// ...
setOpen(!isBlur);
```

When the test triggered `mouseout`, `isBlur` was `false` (since `'mouseout' !== 'mouseleave'`), so `setOpen(true)` was called, keeping the sidebar expanded instead of collapsing it.

## Test Plan

- [ ] Cypress E2E tests pass on CI
- [ ] The `sidebar.cy.ts` hover test specifically passes


amikofalvy:
Closing - fix was pushed directly to main

PR #1227, opened 2025-12-08
multi tenant signoz

shagun-singh-inkeep:


shagun-singh-inkeep:
@claude can you do a review and verify the tenant id that we use for our signoz api requests comes from the validated request context

PR #1232, opened 2025-12-09
Empty agent prompt field not updating in database

Gaurav1998-dev:
## Summary

Fixes a bug where empty string prompts were not being serialized, preventing database updates when clearing the prompt field.

Bug demo:
https://github.com/user-attachments/assets/de4b6928-c651-4110-bd25-814396ae598d

## Problem

The original condition `if (metadata?.prompt)` would skip empty strings since they're falsy in JavaScript. This meant that when a user cleared the prompt field (setting it to an empty string), the field wouldn't be included in the serialized result, preventing the database from updating the prompt to an empty value.

## Solution

By explicitly checking for empty strings (`metadata?.prompt === ''`), we ensure that empty prompts are still serialized and sent to the database, allowing users to clear the prompt field.

## Testing

- Verified that empty string prompts are now properly serialized
- Verified that non-empty string prompts continue to work as expected

Fix demo:
https://github.com/user-attachments/assets/cab2c537-a1e1-475d-a5b5-ab3fdea86512

### DIFF
--- a/.changeset/good-moons-stare.md
+++ b/.changeset/good-moons-stare.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+fix empty agent prompt field not updating in database


agents-manage-ui/src/features/agent/domain/serialize.ts
```diff
@@ -652,7 +652,11 @@ export function serializeAgentData(
     (result as any).stopWhen = metadata.stopWhen;
   }
 
-  if (metadata?.prompt) {
+  /**
+  * Explicitly check for empty string (falsy) to ensure prompt updates in the database.
+  * Without this check, empty strings would be skipped, leaving prompt undefined.
+  */
+  if (metadata?.prompt || metadata?.prompt === '') {
```
dimaMachina:
```suggestion
  if (metadata) {
```

no need comment

dimaMachina (approved):
lgtm

PR #1234, opened 2025-12-09
migrate `postinstall` and `upgrade` scripts to `ts` in `create-agents-template`

dimaMachina:


create-agents-template/scripts/postinstall.ts
```diff
@@ -8,13 +8,11 @@ if (isVercel && !skip) {
   // Export the UI
   execSync('inkeep dev --export --output-dir ./apps/manage-ui', {
     stdio: 'inherit',
-    shell: true,
```
dimaMachina, line 11: image

PR #1235, opened 2025-12-09
Fix generate render with anthropic models

sarah-inkeep:
Summary of issue: 
The problem was z.any() type for mockData in the renderSchema. When this Zod schema gets converted to JSON Schema for Anthropic's structured output API:
z.string() → { "type": "string", "description": "..." } ✅
z.any() → { "description": "Sample data matching the props schema" } ❌ No type field!
Anthropic's API strictly requires all schema properties to have a type field, while OpenAI is more lenient and accepts schemas without explicit types.

### DIFF
--- a/agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts
+++ b/agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts
@@ -9,65 +9,12 @@
  * 4. Streams NDJSON response back to client
  */
 
-import { ModelFactory } from '@inkeep/agents-core';
+import { jsonSchemaToZod, ModelFactory } from '@inkeep/agents-core';
 import { streamObject } from 'ai';
 import type { NextRequest } from 'next/server';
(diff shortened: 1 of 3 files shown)


agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts
```diff
@@ -16,6 +16,59 @@ import { z } from 'zod';
 import { fetchDataComponent } from '@/lib/api/data-components';
 import { fetchProject } from '@/lib/api/projects';
 
+/**
+ * Converts JSON Schema to Zod schema with Anthropic-compatible types.
+ * Avoids z.any() and z.unknown() which don't produce proper JSON Schema types.
+ */
+function jsonSchemaToZod(jsonSchema: Record<string, unknown> | null): z.ZodType<unknown> {
```
Comment (amikofalvy) [inline agents-manage-ui/src/app/api/data-components/[dataComponentId]/generate-render/route.ts]: There are several functions defined in our codebase jsonSchemaToZod. Can we define it in the core package and export it from their for use in all the other packages?

PR #1236, opened 2025-12-09
warnings for tool calls and mcp tool call breakdown

shagun-singh-inkeep:


robert-inkeep:
@claude

```

  // Memoize first error ID for auto-scroll functionality
  const firstErrorId = useMemo(() => {
    const errorActivity = sortedActivities.find(
      (activity) => activity.status === ACTIVITY_STATUS.ERROR || activity.hasError === true
    );
    return errorActivity?.id ?? null;
  }, [sortedActivities]);

  // Ref to track if we've already scrolled to the first error
  const hasScrolledToErrorRef = useRef<string | undefined>(undefined);
  const scrollContainerRef = useRef<HTMLDivElement>(null);
  ```
  
  there's got to be a scroll way to do this right?
  
  a backend state for this sounds wrong

robert-inkeep:
@claude i didnt mean literally backend

i meant literal state variables for this

shagun-singh-inkeep:
@claude can i instead just query for the first error item since querySelector automatically selects the first one??

robert-inkeep:
@claude 

re do the review for this pr

PR #1263, opened 2025-12-12
Midgeneration Compression

tim-inkeep:
Core Features Added:

  1. Mid-Generation Compression System - Automatically compresses conversation context during agent generation to prevent hitting token limits
  2. Manual Compression Tool - Agents can now manually trigger compression with the compress_context tool when shifting topics or completing major tasks
  3. Layered Compression Strategy - Preserves original messages while compressing only generated content (tool results, intermediate steps)
  4. Compression-Aware Artifacts - Enhanced artifact naming with better uniqueness handling and context awareness
  5. Conversation Distillation - New distillation tool that summarizes conversations while preserving key decisions and artifacts

  Technical Improvements:
  - Smart Context Management - Tracks original vs generated messages separately
  - Compression State Tracking - Monitors when compression is needed based on token limits
  - Enhanced Artifact Service - Better handling of summary data and metadata
  - Improved Error Handling - More robust fallback strategies for artifact saving

  New Environment Constants to Watch:
  - AGENTS_COMPRESSION_HARD_LIMIT - Hard token limit (default: 120,000)
  - AGENTS_COMPRESSION_SAFETY_BUFFER - Safety buffer before hitting limit (default: 20,000)
  - AGENTS_COMPRESSION_ENABLED - Toggle compression on/off (default: true)

This enables agents to handle much longer conversations and complex multi-step tasks without running into context length limitations.

### DIFF
--- a/agents-manage-ui/src/components/data-components/render/component-render-generator.tsx
+++ b/agents-manage-ui/src/components/data-components/render/component-render-generator.tsx
@@ -1,18 +1,22 @@
 'use client';
 
-import { Loader2, RefreshCw, Sparkles, Trash2 } from 'lucide-react';
+import { Info, Loader2, RefreshCw, Sparkles, Trash2 } from 'lucide-react';
 import { useCallback, useMemo, useState } from 'react';
 import { toast } from 'sonner';
 import { Streamdown } from 'streamdown';
 import { CodeEditor } from '@/components/editors/code-editor';
 import { JsonEditor } from '@/components/editors/json-editor';
+import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import { Card } from '@/components/ui/card';
+import { ExternalLink } from '@/components/ui/external-link';
 import { InfoCard } from '@/components/ui/info-card';
 import { Label } from '@/components/ui/label';
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover';
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs';
 import { Textarea } from '@/components/ui/textarea';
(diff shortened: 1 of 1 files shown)


agents-run-api/src/agents/Agent.ts
```diff
@@ -377,6 +382,38 @@ export class Agent {
     };
   }
 
+  /**
+   * Get the model settings for summarization/distillation
+   * Falls back to base model if summarizer not configured
+   */
+  private getSummarizerModel(): ModelSettings {
+    if (!this.config.models) {
```
nick-inkeep, line 390: Not familiar with codebase, but since we require base at the project level, shouldn't models.base and this.config.models statically defined as required already?

agents-run-api/src/services/AgentSession.ts
```diff
@@ -169,6 +174,15 @@ export interface ToolResultData {
   needsApproval?: boolean;
 }
 
+export interface CompressionEventData {
+  reason: 'manual' | 'automatic';
+  messageCount: number;
+  artifactCount: number;
+  contextSizeBefore: number;
+  contextSizeAfter: number;
+  compressionType: 'mid_generation' | 'full_conversation';
```
nick-inkeep, line 183: minor nit: claude code uses 'compact' as the term, instead of compression.

nick-inkeep (commented):
Conceptually seems sensible. Seems low risk given it's internal logic only and not easy for users to take a dependency on.

tim-inkeep:
# TODOs Left Unfinished

- Documentation Updates
- Tests
- Unbounded Artifact Growth (should probably drop old ones especially old compressed ones)
- Unbounded Conversation Growth
- Move away from env variables for control

PR #1280, opened 2025-12-15
replace `monaco-editor` with TipTap editor for `Prompt` and `Agent prompt` inputs fields

dimaMachina:


sarah-inkeep:
Could we also set a maxHeight on the variables dropdown, I noticed it can sometimes overflow the page if there are a lot:
Screenshot 2025-12-15 at 12 17 21 PM

sarah-inkeep:
also just noticed a few bugs with expand mode:
1. if I click "Add variable" in the expand mode, it adds the curly bracket but I don't see the dropdown picker.
Screenshot 2025-12-15 at 12 27 54 PM

2. If I try to add a variable after having been in the expand mode, nothing happens https://www.loom.com/share/9f48f6909df94dd28266d2943ef56140

I notice that the prompt in the fact-agent graph has backticks but I don't think they are intended to be rendered as code here since it is part of the instruction to the LLM to include backticks when providing code.
Screenshot 2025-12-15 at 11 46 17 AM

dimaMachina:
> Could we also set a maxHeight on the variables dropdown, I noticed it can sometimes overflow the page if there are a lot: Screenshot 2025-12-15 at 12 17 21 PM

yep, I finally found how to use shadcn dropdown and set `max-h-64` 🙏

PR #1281, opened 2025-12-15
Refresh agent state on save

sarah-inkeep:


tim-inkeep:
sorry i misclicked here @sarah-inkeep

PR #1308, opened 2025-12-17
Lazy-load Monaco Editor

dimaMachina:


agents-manage-ui/src/lib/monaco-editor/monaco-utils.ts
```diff
@@ -3,9 +3,7 @@ import type * as Monaco from 'monaco-editor';
 // Function to check if a token should show a copy icon
 function shouldShowCopyIcon(tokenType: string): boolean {
   switch (tokenType) {
-    case 'string.value.json':
-    case 'number.json':
-    case 'keyword.json':
+    case 'constant':
```
amikofalvy, line 6: @dimaMachina Is this related to lazy load, is it required for the package upgrades or is it auxiliary?

It's fine to keep this in this PR in any case, but I would like to know the reason to make sure this is desired behavior.

dimaMachina: it's for unit test of copy field value feature in readonly json editor

tests pass, it's safe

https://github.com/inkeep/agents/blob/4525c03fea7cbc5ec819163d373a158e6676df94/agents-manage-ui/src/lib/__tests__/monaco-utils.test.tsx#L115-L132

amikofalvy (approved):

PR #1368, opened 2025-12-29
Change body background to sidebar color

dimaMachina:
just try to overscroll

## before

| dark  | light   |
|---|---|
| image  |  image |

## after

| dark  | light   |
|---|---|
|  image |   image|


sarah-inkeep (changes requested):
hm I think this will change the background color for all pages though, for example the login / onboarding flow pages should have the main background color I think.

PR #1424, opened 2026-01-12
Add `dev` watch scripts and skip `d.ts` generation in watch mode across packages

dimaMachina:
I noticed that rebuild isn't triggered for packages on `dev` mode. We need to add `dev` script with watch mode but skip `d.ts` generation to improve `rebuild` speed

## without `dts`

image

## with `dts`

image

### DIFF
--- a/agents-docs/src/components/expandable-row.tsx
+++ b/agents-docs/src/components/expandable-row.tsx
@@ -1,7 +1,7 @@
 'use client';
 
-import { useState } from 'react';
 import type { ReactNode } from 'react';
+import { useState } from 'react';
 
 interface ExpandableRowProps {
   feature: string;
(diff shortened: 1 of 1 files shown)


agents-cli/tsdown.config.ts
```diff
@@ -1,11 +1,10 @@
...
 
-export default defineConfig({
-  watch: process.env.MODE === 'watch',
+export default defineConfig((options) => ({
   entry: ['src/**/*.ts', '!**/__tests__', '!**/*.test.ts'],
-  format: ['esm'],
+  format: 'esm',
   target: 'node20',
-  dts: process.env.MODE !== 'watch',
+  dts: !options.watch,
```
amikofalvy, line 7: Why remove dts generation for dev mode? type declarations are useful when developing yes?

dimaMachina: it's useful but it makes worse the `build` time, let's discuss today

PR #1425, opened 2026-01-12
New tools

anubra266:


agents-run-api/src/utils/stream-helpers.ts
```diff
@@ -270,7 +314,7 @@ export class VercelDataStreamHelper implements StreamHelper {
   }
 
   setSessionId(sessionId: string): void {
-    this.sessionId = sessionId;
+    void sessionId;
```
dimaMachina:
```suggestion
    this.sessionId = sessionId;

File: agents-run-api/src/handlers/executionHandler.ts [hunk 2]
```diff
@@ -239,6 +254,255 @@ export class ExecutionHandler {
...
+            }
+          }
+
+          await sseHelper.complete();
+          await agentSessionManager.endSession(requestId);
+          unregisterStreamHelper(requestId);
+          return { success: true, iterations: 1 };
+        }
+
+        if (resolvedRoutingMessage) {
```
tim-inkeep, line 423: We should have the logic for building the message here and then use the A2A client and the while loop starting at 506 to send the message. we shouldnt make more logic to send messages

PR #1458, opened 2026-01-16
Agents mono api

miles-kt-inkeep:


agents-api/.well-known/workflow/v1/flow.cjs
shagun-singh-inkeep, line 1: super small but the gitignore currently has 

# Generated workflow bundles
agents-eval-api/.well-known/

# Workflow runtime data
agents-eval-api/.workflow-data/

so just edit the file when you merge so the workflow data doesnt get merged

miles-kt-inkeep: good catch thanks

PR #1477, opened 2026-01-18
remove `forwardRef` usages. In React 19, `forwardRef` is no longer necessary. Pass `ref` as a prop instead.

dimaMachina:


dimaMachina:
@claude reverify

PR #1496, opened 2026-01-21
feat: Add multi-part message format for trigger webhooks

amikofalvy:
## Summary

This PR enhances trigger webhook messages to include both text and structured data parts, providing richer context for agents processing trigger events.

### Core Schema & Validation (`@inkeep/agents-core`)
- Make `messageTemplate` optional in the database schema
- Update validation schemas to allow null/undefined messageTemplate
- Add database migration for schema change

### Webhook Handler (`@inkeep/agents-run-api`)
- Messages now use a multi-part format with `parts` array
- When `messageTemplate` is provided: includes both text part (interpolated template) and data part (transformed payload)
- When `messageTemplate` is not provided: includes only the data part
- Data part is always included when `transformedPayload` is not null/undefined (even for empty objects)

### UI (`@inkeep/agents-manage-ui`)
- Update trigger form to make messageTemplate optional
- Improve help text explaining that structured data is always included

### Documentation
- Update triggers documentation with new message format examples
- Document the optional nature of messageTemplate

## Testing

All 27 webhook tests pass, including 4 new tests specifically for multi-part message behavior:
- `should create message with both text and data parts when messageTemplate is provided`
- `should create message with only data part when messageTemplate is not provided`
- `should include data part even for empty object payload`
- `should include transformed payload in data part when outputTransform is configured`

## Changesets

- `@inkeep/agents-core`: minor
- `@inkeep/agents-run-api`: minor  
- `@inkeep/agents-manage-ui`: patch

### DIFF
--- a/.changeset/big-dancers-show.md
+++ b/.changeset/big-dancers-show.md
@@ -2,9 +2,8 @@
 "@inkeep/agents-core": patch
 "@inkeep/agents-sdk": patch
 "@inkeep/agents-manage-ui": patch
-"@inkeep/agents-run-api": patch
 "@inkeep/agents-cli": patch
-"@inkeep/agents-manage-api": patch
+"@inkeep/agents-api": patch
 "@inkeep/agents-manage-mcp": patch
 "@inkeep/ai-sdk-provider": patch
 "@inkeep/create-agents": patch
--- a/.changeset/brave-hats-warn.md
+++ b/.changeset/brave-hats-warn.md
@@ -1,11 +1,9 @@
 ---
 "@inkeep/agents-core": patch
 "@inkeep/agents-sdk": patch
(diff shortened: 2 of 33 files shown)


packages/agents-core/src/constants/otel-attributes.ts
```diff
@@ -83,6 +83,8 @@ export const SPAN_KEYS = {
 
   // Message attributes
   MESSAGE_CONTENT: 'message.content',
+  MESSAGE_PARTS: 'message.parts',
+  MESSAGE_PARTS_COUNT: 'message.parts.count',
```
robert-inkeep, line 87: @amikofalvy let's remove `MESSAGE_PARTS_COUNT`

all the aggregations / computed data can be moved to query side

PR #1503, opened 2026-01-21
fix(agents-core): avoid calling 2 times `getCredentialReference` or `getUserScopedCredentialReference`

dimaMachina:


dimaMachina:
@claude verify it, this function seems used only in agents-api/src/domains/manage/routes/tools.ts where we already calls `getCredentialReference` and `getUserScopedCredentialReference `

PR #1504, opened 2026-01-21
Consolidate to single-phase generation

tim-inkeep:
Simplifies agent generation from 2-phase (thinking + structured output) to single-phase with inline structured output.

  Key changes:
  - Removed Phase 2 config, templates, and generation logic
  - Merged thinking and structured output into single LLM call
  - Simplified Agent.ts generation flow (~450 lines removed)
  - Updated model selection to use structuredOutput model when data components present

### DIFF
--- a/agents-docs/_snippets/multi-agent-framework/data-components-model-recommendations.mdx
+++ b/agents-docs/_snippets/multi-agent-framework/data-components-model-recommendations.mdx
@@ -1,0 +1,17 @@
+## Model Recommendations
+
+<Warning>
+**Data components require high-capability models for best results.**
+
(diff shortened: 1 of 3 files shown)


.changeset/violet-feet-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-api": patch
+---
+
+Updated Api
```
amikofalvy: changelog message should have more detail.

amikofalvy:
- removal of phase2 static files like xml, .ts files that are no longer used.
- adding docs for config etc

PR #1553, opened 2026-01-22
Add capablities schema descriptions

anubra266:


agents-api/src/createApp.ts
```diff
@@ -51,12 +51,24 @@ function createAgentsHono(config: AppConfig) {
...
+          .enum(['native', 'vercel'])
+          .optional()
+          .describe('The configured sandbox provider, if enabled.'),
+        runtime: z
+          .enum(['node22', 'typescript'])
+          .optional()
+          .describe('The configured sandbox runtime, if enabled.'),
+      })
+      .describe('Sandbox execution capabilities (used by Function Tools).'),
+  }).describe('Optional server capabilities and configuration.');
```
amikofalvy: There also need to be some openapi metadata so we can register this schema for re-use

amikofalvy (approved):

PR #1587, opened 2026-01-26
feat(docs): use latest openapi docs, split api reference docs per tags

dimaMachina:


agents-docs/package.json
```diff
@@ -36,13 +36,13 @@
     "clsx": "^2.1.1",
     "fumadocs-core": "~16.1.0",
     "fumadocs-mdx": "~14.0.3",
-    "fumadocs-openapi": "~10.0.11",
+    "fumadocs-openapi": "^10.2.6",
```
amikofalvy: Fumadocs doesn't follow semver exactly. They often publish breaking changes under minor versions. quite easily. please leave `~` in place.

agents-docs/scripts/generate-openapi-docs.ts
```diff
@@ -1,21 +1,83 @@
...
-    includeDescription: true,
+    per: 'tag',
+    frontmatter(title) {
+      return {
+        title: title
+          .replace('A2 A', 'A2A')
+          .replace('A P I', 'API')
+          .replace('C L I', 'CLI')
+          .replace('O Auth', 'OAuth')
+          .replace('M C P', 'MCP'),
```
amikofalvy, line 74: What is the purpose of these replacements? Shouldn't the titles be correct in the first place?

agents-docs/package.json
dimaMachina: replaced

agents-docs/scripts/generate-openapi-docs.ts
dimaMachina: added a comment

PR #1588, opened 2026-01-26
Create new agents-mcp package

amikofalvy:


.github/workflows/ci.yml
```diff
@@ -202,6 +202,7 @@ jobs:
           # These need dist/ directories for subpath exports like ./factory
           pnpm --filter @inkeep/agents-core build
```
dimaMachina: we should use turbo here
```suggestion
          turbo run build --filter=@inkeep/create-agents
```

PR #1593, opened 2026-01-27
chore(api): reduce openapi tags

dimaMachina:


dimaMachina:
@claude based on my refactoring

Send a proposal for other tags in https://github.com/inkeep/agents/pull/1593/files#diff-3f7dc20d1b7d3e74babe2452dd13d69cb091bee0d6dbb9688da0337613488a06

Which one we should remove/rename

agents-api/__snapshots__/openapi.json
```diff
@@ -13213,7 +13213,8 @@
         },
         "summary": "Associate Artifact Component with Agent",
         "tags": [
-          "Agent Artifact Component Relations"
+          "Agents",
+          "Artifact Components"
```
amikofalvy, line 13217: Doesn't this add confusion on where to find endpoint documentation?


User story: I'm told to use the Associate Artifact Component with Agent endpoint, I go to look for it in the documentation. "Where do I look for it"? I see some relational entities have their own side bar navigation item

Image

We should at least be consistent in how we show nav for relational entities

amikofalvy: Remove 'Agent Tool Relations' → use 'Agents' & 'Tools'
Remove 'SubAgent Team Agent Relations' → use 'Sub Agents'

PR #1596, opened 2026-01-27
GitHub app management

miles-kt-inkeep:


agents-manage-ui/src/components/settings/settings-nav.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const navItems = [
+  {
+    label: 'Organization',
+    href: (tenantId: string) => `/${tenantId}/settings`,
+    icon: Settings,
+    exact: true,
+  },
+];
```
sarah-inkeep: are we planning on adding more items here? I guess my thought is that if there is only one tab maybe we don't need the tab at all?

agents-manage-ui/src/components/settings/settings-nav.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function SettingsNav({ tenantId }: SettingsNavProps) {
```
sarah-inkeep: this code looks very similar if not identical to WorkAppsNav maybe we can extract it to be a generic component that takes in an array of navItems instead?

agents-manage-ui/src/app/[tenantId]/settings/layout.tsx
```diff
@@ -1,15 +1,19 @@
 import type { Metadata } from 'next';
 import { PageHeader } from '@/components/layout/page-header';
+import { SettingsNav } from '@/components/settings/settings-nav';
 
 export const metadata = {
-  title: 'Organization Settings',
-  description: 'Manage your organization settings',
+  title: 'Settings',
+  description: 'Manage your organization settings and integrations',
```
Comment (sarah-inkeep) [inline agents-manage-ui/src/app/[tenantId]/settings/layout.tsx]: are we planning to add integration settings here? I didn't see any when I pulled the branch but maybe I miss something?

agents-manage-ui/src/app/[tenantId]/work-apps/github/[installationId]/page.tsx
```diff
@@ -0,0 +1,352 @@
...
+  return new Date(dateString).toLocaleDateString('en-US', {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+    hour: '2-digit',
+    minute: '2-digit',
+  });
+}
+
+function formatShortDate(dateString: string) {
```
Comment (sarah-inkeep) [inline agents-manage-ui/src/app/[tenantId]/work-apps/github/[installationId]/page.tsx]: I think we might already have a formatDate function in lib/utils/format-date that we could maybe reuse here

agents-manage-ui/src/components/settings/settings-nav.tsx
miles-kt-inkeep: Oops yeah for all these I moved everything to /work-apps and forgot to delete these settings page changes.

.changeset/awake-white-turkey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-api": patch
+---
+
+agents-core
```
amikofalvy: Need a better changeset entry message here.

agents-api/src/domains/manage/routes/github.ts
```diff
@@ -0,0 +1,713 @@
...
+          },
+        },
+      },
+      ...commonGetErrorResponses,
+    },
+  }),
+  async (c) => {
+    const { tenantId, installationId } = c.req.valid('param');
+
+    logger.info({ tenantId, installationId }, 'Getting GitHub App installation details');
```
amikofalvy, line 242: minor - I'm hoping to get rid of this level of granularity log at the info level. We already have request logging at the info level which includes non-sensitive path parameters. I think we can remove this type of log in the future. @claude make an agent skill that defines when it is appropriate to use logger in api routes. Open in a new PR for review, make the base branch main so as to only include the new skill addition against the main branch.

agents-api/src/domains/manage/routes/github.ts
```diff
@@ -0,0 +1,713 @@
...
+
+    const installation = await getInstallationById(runDbClient)({
```
amikofalvy: We need to be more diligent about reducing sequential db calls. Can we do this query in parallel with the other [getInstallationById, getRepositoriesByInstallationId]? We can resolve the parallel execution if getInstallationById returns none or if getRepositoriesByInstallationId returns >0 count. I think we have all the information necessary to return the right error messages for either of these resolutions.

agents-api/src/domains/manage/routes/github.ts
```diff
@@ -0,0 +1,713 @@
...
+import runDbClient from '../../../data/db/runDbClient';
+import { getLogger } from '../../../logger';
+import type { ManageAppVariables } from '../../../types/app';
+
+const logger = getLogger('github-manage');
+
+const app = new OpenAPIHono<{ Variables: ManageAppVariables }>();
+
+const InstallUrlResponseSchema = z.object({
+  url: z.url().describe('GitHub App installation URL with signed state parameter'),
```
amikofalvy, line 38: nit- does z.url already have a string length restriction? if not we should add it to protect our db

packages/agents-core/src/db/manage/manage-schema.ts
```diff
@@ -344,6 +344,7 @@ export const tools = pgTable(
     imageUrl: text('image_url'),
     capabilities: jsonb('capabilities').$type<ToolServerCapabilities>(),
     lastError: text('last_error'),
+    isWorkapp: boolean('is_workapp').notNull().default(false),
```
amikofalvy: is the concept of work app well documented? I'm a bit confused by the name choice myself so I don't think it is a self evident name.

additionally, is it `workApp` or `workapp`?

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +494,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+
+/**
+ * Tracks GitHub App installations linked to tenants.
+ * One tenant can have multiple installations (e.g., multiple orgs).
+ * The installation_id is the GitHub-assigned ID, unique across all GitHub.
+ */
+export const workAppGitHubInstallations = pgTable(
+  'work_app_github_installations',
+  {
+    id: varchar('id', { length: 256 }).primaryKey(),
```
amikofalvy: improvement - can we make a re-useable definition for id? Since this is the inkeep generated id, we should define a central id schema for re-use across all of our tables. This way we have a standard convention for our primary keys. @claude can you open a PR with this change?

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +494,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+      table.installationDbId,
+      table.repositoryId
+    ),
+    foreignKey({
+      columns: [table.installationDbId],
+      foreignColumns: [workAppGitHubInstallations.id],
+      name: 'work_app_github_repositories_installation_fk',
+    }).onDelete('cascade'),
+  ]
+);
```
amikofalvy, line 563: So this is essentially a caching layer that syncs via github webhooks and we treat this caching layer as our source of truth, yes?

packages/agents-core/src/db/runtime/runtime-schema.ts
miles-kt-inkeep: @claude remove the change from this PR, open it in a new PR against main.

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +465,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+ * The tenant_id and project_id reference the projects table in the manage schema
+ * (cross-schema, no FK constraint for project). tenant_id is included because
+ * project IDs are only unique within a tenant.
+ */
+export const workAppGitHubProjectRepositoryAccess = pgTable(
+  'work_app_github_project_repository_access',
+  {
+    id: inkeepId().primaryKey(),
+    tenantId: tenantId(),
+    projectId: inkeepId('project_id'),
```
amikofalvy: can't we use projectScoped common here?

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +465,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+/**
+ * Tracks GitHub App installations linked to tenants.
+ * One tenant can have multiple installations (e.g., multiple orgs).
+ * The installation_id is the GitHub-assigned ID, unique across all GitHub.
+ */
+export const workAppGitHubInstallations = pgTable(
+  'work_app_github_installations',
+  {
+    id: inkeepId().primaryKey(),
+    tenantId: tenantId(),
```
amikofalvy: tenantScoped here?

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +465,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+ * (cross-schema, no FK constraint). These are denormalized here so all GitHub access
+ * info can be queried from PostgreSQL alone.
+ */
+export const workAppGitHubMcpToolRepositoryAccess = pgTable(
+  'work_app_github_mcp_tool_repository_access',
+  {
+    id: inkeepId().primaryKey(),
+    toolId: inkeepId('tool_id'),
+    tenantId: tenantId(),
+    projectId: inkeepId('project_id'),
```
amikofalvy: projectScoped?

packages/agents-core/src/db/runtime/runtime-schema.ts
```diff
@@ -491,3 +465,241 @@ export const ledgerArtifactsRelations = relations(ledgerArtifacts, ({ one }) =>
...
+  ]
+);
+
+/**
+ * Stores the explicit access mode for project-level GitHub repository access.
+ * - 'all': Project has access to all repositories from tenant GitHub installations
+ * - 'selected': Project only has access to repositories listed in work_app_github_project_repository_access
+ * If no row exists for a project, defaults to 'selected' (fail-safe: no access unless explicitly granted).
+ */
+export const workAppGitHubProjectAccessMode = pgTable(
```
amikofalvy, line 643: Why is this its own table? can't we just make it a column on the project? only unique thing here is the `mode`

packages/agents-core/src/types/utility.ts
```diff
@@ -389,3 +389,19 @@ export type DatasetItemInput = {
...
 export type DatasetItemExpectedOutput = Array<{ role: string; content: MessageContent }>;
+
+/**
+ * GitHub App installation status.
+ * - 'pending': Installation request awaiting org admin approval
+ * - 'active': Installation is active and functional
+ * - 'suspended': Installation suspended by GitHub or org admin
+ * - 'disconnected': Installation has been disconnected by the user (soft delete)
+ */
+export type WorkAppGitHubInstallationStatus = 'pending' | 'active' | 'suspended' | 'disconnected';
```
amikofalvy: mirrors enums from github?

amikofalvy: make note if so.

packages/agents-core/src/validation/schemas.ts
```diff
@@ -2531,3 +2535,67 @@ export const ProjectMetadataSelectSchema = createSelectSchema(projectMetadata);
...
+
+export const WorkAppGitHubInstallationSelectSchema = createSelectSchema(workAppGitHubInstallations);
+export const WorkAppGitHubInstallationInsertSchema = createInsertSchema(workAppGitHubInstallations)
+  .omit({
+    createdAt: true,
+    updatedAt: true,
+    status: true,
+  })
+  .extend({
+    accountType: z.enum(['Organization', 'User']),
```
amikofalvy: we are redefining the schema here when we have already defined them in packages/agents-core/src/types/utility.ts

amikofalvy: why don't we just define the zod schema and extract a type from the zod schema in packages/agents-core/src/types/utility.ts. We can reference the zod schema directly in this extension.

packages/agents-core/src/validation/schemas.ts
```diff
@@ -2531,3 +2535,67 @@ export const ProjectMetadataSelectSchema = createSelectSchema(projectMetadata);
...
+    status: z.enum(['pending', 'active', 'suspended', 'disconnected']).optional().default('active'),
+  });
+
+export const WorkAppGithubInstallationApiSelectSchema = WorkAppGitHubInstallationSelectSchema.omit({
+  tenantId: true,
```
amikofalvy: I think we have schema helper utilities already defined to omit common properties like tenantId and even id or createdAt/updatedAt as we are replicating below.

packages/agents-core/src/env.ts
```diff
@@ -55,6 +55,7 @@ const envSchema = z.object({
   BETTER_AUTH_SECRET: z.string().optional(),
   TRUSTED_ORIGIN: z.string().optional(), // for local/preview env
   OAUTH_PROXY_PRODUCTION_URL: z.string().optional(), // for local/preview env
+  GITHUB_MCP_API_KEY: z.string().optional(),
```
amikofalvy, line 58: just a thought - it may make sense to add descriptions to these schemas to elaborate on what each of these env vars is useful for.

packages/agents-work-apps/src/github/mcp/auth.ts
```diff
@@ -0,0 +1,33 @@
...
+import { env } from '../../env';
+
+export const githubMcpAuth = () =>
+  createMiddleware<{
+    Variables: {
+      toolId: string;
+    };
+  }>(async (c, next) => {
+    const toolId = c.req.header('x-inkeep-tool-id');
+    if (!toolId) {
```
amikofalvy, line 12: These error responses are simple and they will work. I would recommend adopting the Application JSON Problem Details spec that we use in our other apis to make explicit human/machine readable error responses that support a schema for highlighting which parameter/ path/header/body property is causing the error.

packages/agents-work-apps/src/github/mcp/index.ts
```diff
@@ -0,0 +1,1079 @@
...
+  generatePrMarkdown,
+  getGitHubClientFromRepo,
+  type LLMUpdateOperation,
+  visualizeUpdateOperations,
+} from './utils';
+
+const updateOperationSchema = z
+  .object({
+    operation: z
+      .enum(['replace_lines', 'insert_after', 'insert_before', 'delete_lines'])
```
amikofalvy, line 29: another duplicate schema for operation. Please re-use operation Schema that we talked about elsewhere in the gh comments.

packages/agents-work-apps/src/github/mcp/index.ts
```diff
@@ -0,0 +1,1079 @@
...
+    `Find files in a repository using GitHub's search API. ${getAvailableRepositoryString(repositoryAccess)}`,
+    {
+      owner: z.string().describe('The owner of the repository'),
+      repo: z.string().describe('The name of the repository'),
+      query: z.string().describe('The query to search for'),
+      path: z
+        .string()
+        .optional()
+        .describe('The path to the files to search in relative to the root of the repository'),
+      limit: z.number().optional().describe('The maximum number of files to return (default: 30)'),
```
amikofalvy: should we add default to the zod schema to make sure the default lines up or should we just continue to use it in the description even though this is disconnected from the actual default parameter value in the definition of the tool callback below `limit = 30`

packages/agents-core/src/db/runtime/runtime-schema.ts
miles-kt-inkeep: Leaving this as is for now. I think this is cleaner than interpreting a nullable value on the projectMetadata table and I want to avoid changing schema this late.

PR #1598, opened 2026-01-27
chore(agents-docs): add `/api/proxy` route to fix `Failed to fetch` in playground OpenAPI pages

dimaMachina:
this will fix broken playground in OpenAPI pages

image

and we'll avoid using workaround with `/api/proxy` by fumadocs recommendation https://www.fumadocs.dev/docs/integrations/openapi/server#setup

### DIFF
--- a/agents-docs/src/app/api/proxy/route.ts
+++ b/agents-docs/src/app/api/proxy/route.ts
@@ -1,7 +1,7 @@
 import { openapi } from '@/lib/openapi';
 
 /**
- * @see https://www.fumadocs.dev/docs/integrations/openapi/server#setup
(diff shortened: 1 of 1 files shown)


agents-docs/src/app/api/proxy/route.ts
```diff
@@ -0,0 +1,9 @@
+import { openapi } from '@/lib/openapi';
+
+/**
+ * @see https://www.fumadocs.dev/docs/integrations/openapi/server#setup
```
dimaMachina:
```suggestion
 * @see https://fumadocs.dev/docs/integrations/openapi/server#setup
```

PR #1634, opened 2026-01-29
fix(agents-core): add AST validation for function tools `executeCode`

dimaMachina:


packages/agents-core/src/validation/schemas.ts
```diff
@@ -1668,8 +1669,67 @@ export const FunctionInsertSchema = createInsertSchema(functions).extend({
...
+    // Workaround for anonymous function because it’s not valid JavaScript grammar.
+    // Babel (and every JS parser) rejects it.
+    const isAnonymousFunction = /^(async\s+)?function(\s+)?\(/.test(val);
+    if (isAnonymousFunction) {
+      val = `(${val})`;
+    }
+    const ast = parse(val, { sourceType: 'module' });
+    const { body } = ast.program;
+    for (const node of body) {
+      if (node.type === 'ExportDefaultDeclaration') {
```
anubra266, line 1684: we don't want any kinda export both default and non default. can we add a message for export too

dimaMachina: added this check

PR #1635, opened 2026-01-29
fix(dashboard): add toast with error when user want to remove undeletable nodes

dimaMachina:
I was confused why my node isn't removed after deleting `Default` node

image

### DIFF
--- a/agents-manage-ui/src/features/agent/state/use-agent-store.ts
+++ b/agents-manage-ui/src/features/agent/state/use-agent-store.ts
@@ -324,7 +324,7 @@ const agentState: StateCreator<AgentState> = (set, get) => ({
         if (unDeletableNodes.length) {
           const formatter = new Intl.ListFormat('en', { type: 'conjunction' });
           toast.error(
-            `Cannot delete undeletable nodes ${formatter.format(unDeletableNodes.map((n) => n.id))}`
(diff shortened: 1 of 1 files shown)


agents-manage-ui/src/features/agent/state/use-agent-store.ts
```diff
@@ -318,6 +319,15 @@ const agentState: StateCreator<AgentState> = (set, get) => ({
         const nodesToDelete = new Set(
           state.nodes.filter((n) => n.selected && (n.deletable ?? true)).map((n) => n.id)
         );
+
+        const unDeletableNodes = state.nodes.filter((n) => n.selected && !n.deletable);
+        if (unDeletableNodes.length) {
+          const formatter = new Intl.ListFormat('en', { type: 'conjunction' });
+          toast.error(
+            `Cannot delete undeletable nodes ${formatter.format(unDeletableNodes.map((n) => n.id))}`
```
sarah-inkeep: what do you think about making this error message something like "Cannot delete default subagent." instead?

dimaMachina:
```suggestion
            `Cannot delete default subagent ${formatter.format(unDeletableNodes.map((n) => n.id))}`
```

PR #1641, opened 2026-01-29
fix(dashboaard): update `react-resizable-panels` and `next` to fix `Previous layout not found for panel index 2` error

dimaMachina:


.changeset/great-suns-see.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+fix(dashboaard): update `react-resizable-panels` and `next` to fix `Previous layout not found for panel index 2` error
```
dimaMachina, line 5:
```suggestion
fix(dashboard): update `react-resizable-panels` and `next` to fix `Previous layout not found for panel index 2` error
```

PR #1671, opened 2026-02-02
fix broken `pnpm --filter @inkeep/agents-manage-ui test:e2e`

dimaMachina:
Load environment variables from .env file.


agents-manage-ui/cypress/support/e2e.ts
```diff
@@ -15,8 +15,8 @@
 
 import './commands';
 
-// Login before each test to ensure authenticated state
-beforeEach(() => {
+// Login before all tests
+before(() => {
```
dimaMachina:
```suggestion
// Login before each test to ensure authenticated state
beforeEach(() => {
```

PR #1699, opened 2026-02-04
feat(dashboard): Notify users to enter custom headers if required in chat or if custom headers are invalid

dimaMachina:
https://inkeep.slack.com/archives/C08QXR5CWBH/p1770217539550219?thread_ts=1770145326.982309&cid=C08QXR5CWBH

### DIFF
--- a/.changeset/custom-headers-validation.md
+++ b/.changeset/custom-headers-validation.md
@@ -1,0 +1,6 @@
+---
+"@inkeep/agents-manage-ui": patch
+"@inkeep/agents-core": patch
+---
+
+Add custom headers validation in playground chat. Users are now notified when custom headers are invalid or required based on the agent's headers schema configuration.


miles-kt-inkeep:
@inkeep can you check for docs changes.

agents-manage-ui/src/components/agent/playground/chat-widget.tsx
```diff
@@ -48,6 +50,20 @@ const styleOverrides = `
 }
 `;
 
+const styleHeadersError = css`
+.ikp-ai-chat-input__fieldset {
+  border: 1px #ef4444 solid;
+  &:after {
+    content: 'Custom headers are invalid.';
+    position: absolute;
```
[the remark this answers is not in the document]
dimaMachina: https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler#annotations

packages/agents-core/src/utils/schema-conversion.ts
```diff
@@ -14,6 +14,9 @@ type PreviewZodType = z.ZodTypeAny & {
 };
 
 /**
+ * TODO @Miles this function treats even non required JSON schema field as required, we should probably use
+ * import { convertJsonSchemaToZod } from 'zod-from-json-schema';
+ *
```
dimaMachina, line 19: cc @miles-kt-inkeep I think we should use `zod-from-json-schema` package, because our own `jsonSchemaToZod` package incorrectly convert json schema and treats non required fields as required

dimaMachina:
> I think we should only show the error, block submission if the field(s) that are required in the headers are missing, but I think the chat still works if the non required headers are not set.
> 
> For example for this config, I don't think we should show any errors:
> 
> ```
> {
>   "type": "object",
>   "required": [],
>   "properties": {
>     "tz": {
>       "type": "string"
>     }
>   },
>   "additionalProperties": false
> }
> ```
> 
> Can you also have Miles review the changes to agents-api / agents-core, I don't think I have enough context there to review properly.

yep, our custom `jsonSchemaToZod` incorrectly convert non required json schema fields as required zod schema fields, see https://github.com/inkeep/agents/pull/1699#discussion_r2767619162

I am using now other package in dashboard, can you reverify?

packages/agents-core/src/utils/__tests__/convert-json-schema-to-zod.test.ts
miles-kt-inkeep, line 1: should this be in `__tests__`?

packages/agents-core/src/utils/schema-conversion.ts
miles-kt-inkeep: Yes lets do it

dimaMachina:
@claude can you write a changeset based on changes in this PR

PR #1712, opened 2026-02-04
bump zod to latest 4.3.6 and fix `.omit() cannot be used on object schemas containing refinements` error

dimaMachina:


packages/create-agents/package.json
```diff
@@ -70,8 +70,5 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
-  },
-  "overrides": {
-    "zod": "^4.1.11"
```
dimaMachina, line 75: looks like we don't need this override

dimaMachina:
@claude verify

PR #1731, opened 2026-02-05
Fix `ModelSelector` layout where custom model inputs were cut off and improve `ModelSelector` architecture with new shadcn `ButtonGroup` component

dimaMachina:


dimaMachina:
@claude add changeset based on all changes and that custom model was cut of previously was fixed

PR #1737, opened 2026-02-05
feat: add image support

mike-inkeep:
- **Add image handling support (without persistence to conversation history)**
- **Tweak image URL schema and tighten up tests**

### DIFF
--- a/agents-api/src/__tests__/run/agents/generateTaskHandler.test.ts
+++ b/agents-api/src/__tests__/run/agents/generateTaskHandler.test.ts
@@ -700,7 +700,7 @@ describe('generateTaskHandler', () => {
       const result = await taskHandler(task);
 
       expect(result.status.state).toBe(TaskState.Failed);
-      expect(result.status.message).toBe('No text content found in task input');
+      expect(result.status.message).toBe('No content found in task input');
       expect(result.artifacts).toEqual([]);
     });
 
--- a/agents-api/src/domains/run/agents/generateTaskHandler.ts
+++ b/agents-api/src/domains/run/agents/generateTaskHandler.ts
@@ -1,7 +1,6 @@
 import {
   type AgentConversationHistoryConfig,
   type CredentialStoreRegistry,
-  type DataPart,
   type FilePart,
   type FullExecutionContext,
   generateId,
@@ -66,31 +65,17 @@ export const createTaskHandler = (
         .map((part) => part.text)
         .join(' ');
 
-      // Extract data parts (e.g., from trigger payloads)
-      const dataParts = task.input.parts.filter(
-        (part): part is DataPart => part.kind === 'data' && part.data != null
+      const hasImages = task.input.parts.some(
+        (part): part is FilePart =>
+          part.kind === 'file' && part.file.mimeType?.startsWith('image/') === true
       );
+      const hasData = task.input.parts.some((p) => p.kind === 'data');
 
-      // Build user message: combine text with any structured data
-      let userMessage = textParts;
-
-      // If there are data parts, append them as structured context for the LLM
-      if (dataParts.length > 0) {
-        const dataContext = dataParts
-          .map((part) => {
-            const metadata = part.metadata as Record<string, unknown> | undefined;
-            const source = metadata?.source ? ` (source: ${metadata.source})` : '';
-            return `\n\n<structured_data${source}>\n${JSON.stringify(part.data, null, 2)}\n</structured_data>`;
-          })
-          .join('');
-        userMessage = `${textParts}${dataContext}`;
-      }
-
-      if (!userMessage.trim()) {
+      if (!textParts.trim() && !hasImages && !hasData) {
         return {
           status: {
             state: TaskState.Failed,
-            message: 'No text content found in task input',
+            message: 'No content found in task input',
           },
           artifacts: [],
         };
@@ -367,7 +352,7 @@ export const createTaskHandler = (
       logger.info({ contextId }, 'Context ID');
       logger.info(
         {
-          userMessage: userMessage.substring(0, 500), // Truncate for logging
+          userMessage: textParts.substring(0, 500), // Truncate for logging
           inputPartsCount: task.input.parts.length,
           textPartsCount: task.input.parts.filter((p) => p.kind === 'text').length,
           dataPartsCount: task.input.parts.filter((p) => p.kind === 'data').length,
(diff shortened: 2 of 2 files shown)


agents-api/src/domains/run/utils/message-parts.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { FilePart, Part, TextPart } from '@inkeep/agents-core';
+import type { ContentItem } from '../types/chat';
+
+export const imageUrlSchema = z.union([
+  z.httpUrl(),
+  z
+    .string()
+    .regex(
+      /^data:image\/(png|jpeg|jpg|webp);base64,/,
+      'Image must be PNG, JPEG, or WebP format (GIF not supported by all providers)'
```
robert-inkeep: can you check this? i believe we can support the gif file format. but it's just that only a single frame will work, it cannot understand animated gifs. and that is ok

mike-inkeep: claude and chatgpt both support GIF, but Gemini only supports these: https://ai.google.dev/gemini-api/docs/image-understanding#supported-formats

```
Gemini supports the following image format MIME types:

PNG - image/png
JPEG - image/jpeg
WEBP - image/webp
HEIC - image/heic
HEIF - image/heif
```

Alternatively, we can accept GIF and let the downstream provider error bubble up if we hit this case

amikofalvy: I think skip gif for now until we figure out how to properly expose agent capabilities to the front end for client side validation.

agents-api/src/domains/run/types/chat.ts
```diff
@@ -1,8 +1,18 @@
-export type ContentItem = {
-  type: string;
-  text?: string;
+export type TextContentItem = {
+  type: 'text';
+  text: string;
 };
 
+export type ImageContentItem = {
+  type: 'image_url';
```
amikofalvy: Are we redefining types and schemas?

https://github.com/inkeep/agents/pull/1737/changes#diff-d4594cf84e549a2289ce15381dea24e93b0a15bfa4211ffd785023abd461b99cR67-R73

I think it's a mistake made throughout our codebase, but we try to reduce schemas and types were we can.

You can infer types from zod schemas.  https://zod.dev/basics?id=inferring-types

agents-api/src/domains/run/agents/Agent.ts
```diff
@@ -104,6 +113,20 @@ export function hasToolCallWithPrefix(prefix: string) {
 
 const logger = getLogger('Agent');
 
+export type UserInput =
+  | string
+  | {
```
amikofalvy: I think at this point we can follow the A2A protocol for parts.

I don't want to introduce too many intermediate types in the message handling.

I think we support: Vercel Streaming, A2A, OpenAI Completions, but our primary pipeline (our widget to our agent framework) does Vercel Streaming (Client Widget to Agents Server) to A2A (Agent Server internal messaging)

amikofalvy: I suggest defining types int he packages where we define A2A types or just re-using the existing by scoping down/omitting properties that we are not supporting in the A2A at this time.

agents-api/src/domains/run/agents/Agent.ts
```diff
@@ -87,6 +88,14 @@ import { toolSessionManager } from './ToolSessionManager';
 import type { SystemPromptV1 } from './types';
 import { Phase1Config, V1_BREAKDOWN_SCHEMA } from './versions/v1/Phase1Config';
 import { Phase2Config } from './versions/v1/Phase2Config';
+
+type AiSdkContentPart = {
+  type: string;
+  text?: string;
+  image?: string | URL;
+  experimental_providerMetadata?: { openai?: { imageDetail?: 'auto' | 'low' | 'high' } };
```
amikofalvy: Even just the enum component can make a useful extraction as a schema so we don't have to redefine the enum set.

miles-kt-inkeep:
@inkeep docs updates?

amikofalvy:
@claude re-review

agents-api/src/domains/run/routes/chatDataStream.ts
```diff
@@ -53,9 +56,17 @@ const chatDataStreamRoute = createRoute({
...
+                        type: z.literal('image'),
+                        text: imageUrlSchema,
+                      }),
                       z.object({
                         type: z.union([
-                          z.enum(['text', 'image', 'audio', 'video', 'file']),
+                          z.enum(['audio', 'video', 'file']),
                           z.string().regex(/^data-/, 'Type must start with "data-"'),
                         ]),
                         text: z.string().optional(),
```
amikofalvy, line 74: @shagun-singh-inkeep added these schema/type definitions in an earlier PR that you can re-use now

https://github.com/inkeep/agents/pull/1789/changes#diff-a6b7ad468243ec559f66f3a1cb18b172f224383d7a2f31a15a6c3657f20103d3

agents-api/src/__tests__/run/agents/Agent.test.ts
```diff
@@ -9,6 +9,8 @@ import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { Agent, type AgentConfig } from '../../../domains/run/agents/Agent';
 import { Phase1Config } from '../../../domains/run/agents/versions/v1/Phase1Config';
 
+const toParts = (text: string) => [{ kind: 'text' as const, text }];
```
amikofalvy: I know it is just a test helper function, but I think function names like toXXXX should be reserved for methods on objects/classes so it is clear subject/verb agreement.

so if you added toParts to the string prototype functions, it would make sense as `'hello'.toParts, but if it is a standalone helper function I prefer the naming convention like the following:
```suggestion
const makeTextPart = (text: string) => [{ kind: 'text' as const, text }];
```

agents-api/src/__tests__/run/agents/generateTaskHandler.test.ts
```diff
@@ -131,7 +131,11 @@ vi.mock('../../../domains/run/agents/Agent.js', () => ({
       // Mock implementation
     }
 
-    async generate(message: string, _options: any) {
+    async generate(parts: any[], _options: any) {
```
amikofalvy: Is this mock correct? I thought the function parameter was called userParts and had a known type.

PR #1738, opened 2026-02-05
open custom headers modal if there is validation errors

dimaMachina:


dimaMachina:
@claude add changeset based on changes in this PR

agents-manage-ui/src/components/agent/playground/custom-headers-dialog.tsx
dimaMachina:
```suggestion
import type { FC } from 'react';
```

PR #1743, opened 2026-02-05
fix: route internal A2A and self-calls in-process to prevent multi-instance stream failures

nick-inkeep:
## Background: How Streaming Works Today

When a user sends a chat message, the system opens a Server-Sent Events (SSE) stream back to the client. The response is generated by an AI agent, which writes text deltas (partial tokens) to this stream in real-time as the model generates.

The component that manages writing to the SSE stream is called a **`StreamHelper`**. Each active request creates a `StreamHelper` and registers it in an in-memory `Map` (the **stream helper registry**), keyed by a unique `requestId`.

When the agent runs, it looks up this `StreamHelper` by `requestId` in the registry and writes text deltas to it. This is how streaming text reaches the user's browser in real time.

## Background: How Internal Agent Calls Work

The system supports multi-agent orchestration: a coordinating **ExecutionHandler** dispatches work to individual agents via the **A2A (Agent-to-Agent) protocol**. This dispatch is implemented as an HTTP call using `A2AClient`, which sends a JSON-RPC request to the service's own `/run/agents/a2a` endpoint.

The call flow looks like this:

```
User Request
  → Route Handler (creates StreamHelper, registers in local Map)
    → ExecutionHandler
      → A2AClient.sendMessage() [HTTP POST to /run/agents/a2a]
        → Middleware (auth, project config, ref resolution, baggage)
          → A2A Route Handler
            → Agent.generate()
              → Agent looks up StreamHelper in registry by requestId
              → Agent writes text deltas to StreamHelper → SSE → User
```

The critical detail: the `A2AClient` call goes **over HTTP**. In a single-instance deployment, this HTTP request loops back to the same process. In a multi-instance deployment (load-balanced containers, serverless functions), **the request can land on a different instance**.

## The Problem

When the A2A HTTP call lands on a **different instance** than the one that registered the `StreamHelper`:

1. The route handler on Instance A creates a `StreamHelper` and stores it in Instance A's in-memory registry.
2. `ExecutionHandler` on Instance A sends an HTTP request to `/run/agents/a2a`.
3. The load balancer routes this to **Instance B**.
4. Instance B's agent runs, generates a response, and looks up the `StreamHelper` in **Instance B's** registry.
5. The `StreamHelper` isn't there — it's in Instance A's memory.
6. **Result**: The agent completes successfully (visible in traces/OTEL), but no text deltas are written to the user's SSE stream. The user sees an empty or prematurely-closed response.

This is **intermittent** because it depends on load balancer routing. When the request happens to land on the same instance, everything works. When it doesn't, streaming silently fails.

### Scope of the Problem

An audit of the codebase identified **5 locations** where the service makes HTTP calls back to itself, all vulnerable to the same class of issue:

| # | Location | Mechanism | Middleware required? |
|---|---|---|---|
| 1 | `executionHandler.ts` — Transfer loop | `A2AClient.sendMessage()` | Yes (auth, project, ref, baggage) |
| 2 | `relationTools.ts` — Internal delegation | `A2AClient.sendMessage()` | Yes |
| 3 | `relationTools.ts` — Team agent delegation | `A2AClient.sendMessage()` | Yes |
| 4 | `EvaluationService.ts` — Evals calling chat API | `fetch()` | Yes (auth, project, ref, baggage) |
| 5 | `createApp.ts` — Workflow queue forwarding | `fetch()` | Currently none, but could change |

Sites 1-3 directly cause the streaming bug. Sites 4-5 are the same architectural anti-pattern (self-call over HTTP) with different symptoms: unnecessary latency, double invocation cost on serverless, and fragility when the service can't reach itself by hostname.

### Design principle

> When calling your own service, use `getInProcessFetch()` instead of `fetch()`.

This is a drop-in replacement for `fetch` that routes requests through the Hono application's full middleware stack **in-process** rather than over the network. It is:

- **Same type as `fetch`** — `typeof fetch`. No custom types, no parallel interfaces.
- **Same middleware** — requests go through auth, project config, ref resolution, and baggage, identically to a real HTTP request.
- **Same-instance guaranteed** — the `StreamHelper` registry lookup always succeeds because the agent runs in the same process that registered it.
- **One module, one function** — all 5 call sites use the same abstraction.

### Implementation

**New module**: `agents-api/src/utils/in-process-fetch.ts`

A minimal module (~10 lines) that provides:
- `registerAppFetch(fn)` — called once at app startup with `app.request.bind(app)`
- `getInProcessFetch()` — returns the in-process fetch, or falls back to global `fetch` if not registered

**`A2AClient` change**: Add an optional `fetchFn` field to `A2AClientOptions`. Internal callers pass `getInProcessFetch()`. External callers omit it (defaults to real `fetch`). The `A2AClient` uses `this.fetchFn` wherever it previously called bare `fetch`.

**Call site changes**: Each of the 5 sites swaps `fetch` for `getInProcessFetch()`:
- Sites 1-3: Pass `fetchFn: getInProcessFetch()` to `A2AClient` constructor
- Site 4: Replace `fetch(chatUrl, ...)` with `getInProcessFetch()(chatUrl, ...)`
- Site 5: Replace `fetch(forwardedRequest)` with `getInProcessFetch()(forwardedRequest)`

### What doesn't change

- `ExecutionHandlerParams` interface — no new fields
- Route handlers (`chat.ts`, `chatDataStream.ts`, `TriggerService.ts`) — no changes
- `@inkeep/agents-core` types — no changes
- `StreamHelper` or `stream-registry.ts` — works as-is since calls are in-process
- Response parsing logic in `ExecutionHandler` (transfers, artifacts, etc.)
- External A2A calls (to external agents) — still use real HTTP `fetch`

### Option: Sticky sessions / session affinity

Route all requests for a given session to the same instance via load balancer configuration.

**Discarded because**: Unsupported on Vercel serverless. Brittle even on platforms that offer it — fails under instance scaling, restarts, and retries. Doesn't solve the fundamental architectural issue.

### Option: Distributed streaming bus (Redis pub/sub, NATS)

Replace the in-memory `StreamHelper` registry with a distributed pub/sub system. Agents publish text deltas to a shared bus; the coordinator subscribes and forwards to the client.

**Discarded because**: Significant infrastructure and complexity overhead (ordering, backpressure, reconnect/resume, new failure modes). Overkill when the root cause is simply "the HTTP call left the process unnecessarily." The in-process fix is simpler, has zero new infrastructure dependencies, and fully solves the problem for single-instance deployments.

### Option: Direct function call (bypass HTTP stack)

Call the agent handler function directly from `ExecutionHandler`, skipping HTTP routing and middleware entirely.

**Discarded because**: This would duplicate or skip middleware logic (auth, project config, ref resolution, baggage). Any future middleware additions would need to be manually replicated in the direct call path, creating silent drift between internal and external request handling.

## Future Considerations: Distributed Execution Mode

The `getInProcessFetch()` abstraction is designed to be the **deployment-mode configuration point** for future distributed architectures.

### Current state: single-instance mode

All internal calls route in-process via `app.request()`. The `StreamHelper` is in the same process as the agent. Streaming works directly.

### Future state: distributed execution mode

In a distributed architecture where a **coordinator** holds the client connection and **workers** execute agents:

1. **Routing** (where does the request go?): `registerAppFetch` would be called with `fetch` (real HTTP to workers via load balancer) or a custom `workerDispatchFetch` (queue-based dispatch). **None of the 5 call sites change.**

2. **Stream bridging** (how do text deltas get back to the coordinator?): The coordinator would consume the worker's streaming HTTP response and pipe it to the local `StreamHelper`. The A2A protocol already supports streaming responses via `sendMessageStream()`, so this is an additive change in `ExecutionHandler`, not a refactor.

The deployment mode would be controlled by an environment variable:

```
# Single-instance (default): internal calls route in-process
EXECUTION_MODE=single

# Distributed: internal calls route over network to workers
EXECUTION_MODE=distributed
```

The key architectural property: **everything below `getInProcessFetch()` is deployment-mode-specific; everything above it is mode-agnostic.** The transition to distributed mode is additive — new env var, new registration logic, new stream piping — not a rewrite.

## Files Changed

| File | Change |
|---|---|
| `agents-api/src/utils/in-process-fetch.ts` | **New** — Module with `registerAppFetch` / `getInProcessFetch` |
| `agents-api/src/domains/run/a2a/client.ts` | Add `fetchFn` to `A2AClientOptions`, use it in `_postRpcRequest` and `_fetchAndCacheAgentCard` |
| `agents-api/src/createApp.ts` | Call `registerAppFetch(base.request.bind(base))` after creating the Hono app |
| `agents-api/src/domains/run/handlers/executionHandler.ts` | Pass `fetchFn: getInProcessFetch()` to `A2AClient` |
| `agents-api/src/domains/run/agents/relationTools.ts` | Pass `fetchFn: getInProcessFetch()` for internal and team delegation |
| `agents-api/src/domains/evals/services/EvaluationService.ts` | Use `getInProcessFetch()` instead of `fetch` for chat API self-call |
| `agents-api/src/createApp.ts` (workflow forward) | Use `getInProcessFetch()` instead of `fetch` for `/index` forwarding |

## Test Plan

- [ ] Existing integration tests pass (they already use `app.request()` via `testRequest.ts`, validating the in-process path)
- [ ] Verify streaming works end-to-end in local dev (single instance)
- [ ] Verify eval dataset runs complete successfully
- [ ] Verify agent transfers produce streamed text in the UI
- [ ] Typecheck passes (`pnpm typecheck`)
- [ ] Lint passes (`pnpm lint`)

### DIFF
--- a/agents-api/src/createApp.ts
+++ b/agents-api/src/createApp.ts
@@ -39,6 +39,7 @@ import { executionBaggageMiddleware } from './middleware/tracing';
 import { setupOpenAPIRoutes } from './openapi';
 import { healthChecksHandler } from './routes/healthChecks';
 import type { AppConfig, AppVariables } from './types';
+import { getInProcessFetch, registerAppFetch } from './utils/in-process-fetch';
 
 const logger = getLogger('agents-api');
 
@@ -345,7 +346,7 @@ function createAgentsHono(config: AppConfig) {
       body: bodyBuffer,
     });
 
-    return fetch(forwardedRequest);
+    return getInProcessFetch()(forwardedRequest);
   });
 
   app.route('/evals', evalRoutes);
@@ -368,6 +369,8 @@ function createAgentsHono(config: AppConfig) {
   // Wrap in base Hono for framework detection
   const base = new Hono();
   base.route('/', app);
+
+  registerAppFetch(base.request.bind(base) as typeof fetch);
 
   return base;
 }
--- a/agents-api/src/domains/evals/services/EvaluationService.ts
+++ b/agents-api/src/domains/evals/services/EvaluationService.ts
@@ -30,6 +30,7 @@ import manageDbPool from '../../../data/db/manageDbPool';
 import runDbClient from '../../../data/db/runDbClient';
 import { env } from '../../../env';
 import { getLogger } from '../../../logger';
+import { getInProcessFetch } from '../../../utils/in-process-fetch';
 
 const logger = getLogger('EvaluationService');
 
@@ -254,7 +255,7 @@ export class EvaluationService {
       'Running dataset item through chat API'
     );
 
-    const response = await fetch(chatUrl, {
+    const response = await getInProcessFetch()(chatUrl, {
       method: 'POST',
       headers,
       body: JSON.stringify(chatPayload),
--- a/agents-api/src/domains/run/a2a/client.ts
+++ b/agents-api/src/domains/run/a2a/client.ts
@@ -48,6 +48,7 @@ export interface A2AClientOptions {
   retryConfig?: RetryConfig;
   ref?: ResolvedRef;
   headers?: Record<string, string>;
+  fetchFn?: typeof fetch;
 }
 
 const DEFAULT_BACKOFF: BackoffStrategy = {
@@ -134,6 +135,7 @@ export class A2AClient {
   private requestIdCounter = 1;
   private serviceEndpointUrl?: string; // To be populated from AgentCard after fetching
   private options: A2AClientOptions;
+  private fetchFn: typeof fetch;
 
   /**
    * Constructs an A2AClient instance.
@@ -145,6 +147,7 @@ export class A2AClient {
    */
   constructor(agentBaseUrl: string, options?: A2AClientOptions) {
     this.agentBaseUrl = agentBaseUrl.replace(/\/$/, ''); // Remove trailing slash if any
+    this.fetchFn = options?.fetchFn ?? fetch;
     this.options = {
       retryConfig: {
         strategy: 'backoff',
@@ -176,7 +179,7 @@ export class A2AClient {
       'agentCardUrl'
     );
     try {
-      const response = await fetch(url.toString(), {
+      const response = await this.fetchFn(url.toString(), {
         headers: {
           Accept: 'application/json',
           ...(this.options.headers || {}),
@@ -221,7 +224,7 @@ export class A2AClient {
         url.searchParams.set('ref', this.options.ref.name);
       }
 
-      const response = await fetch(url.toString(), {
+      const response = await this.fetchFn(url.toString(), {
         headers: {
           Accept: 'application/json',
           ...(this.options.headers || {}),
@@ -465,11 +468,11 @@ export class A2AClient {
     };
 
     const httpResponse = await this.retry(async () => {
-      return fetch(endpoint, {
+      return this.fetchFn(endpoint, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
-          Accept: 'application/json', // Expect JSON response for non-streaming requests
+          Accept: 'application/json',
           ...(this.options.headers || {}),
         },
         body: JSON.stringify(rpcRequest),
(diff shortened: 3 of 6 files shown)


nick-inkeep:
Superseded by #1745 — clean branch with only the in-process fetch transport changes.

PR #1769, opened 2026-02-06
keep both dependencies without hashes and hashed in dashboard

dimaMachina:
Removed the line that deletes the dependency without a hash and updated the dependency addition logic.


dimaMachina:
@claude add changeset based on changes

PR #1799, opened 2026-02-07
feat(dashboard): refactor agent page to use `react-hook-form` and zod schemas from `agents-core`

dimaMachina:
json editors now uses GenericJsonEditor and have expand button too
image

### DIFF
--- a/agents-manage-ui/src/components/agent/sidepane/metadata/metadata-editor.tsx
+++ b/agents-manage-ui/src/components/agent/sidepane/metadata/metadata-editor.tsx
@@ -20,13 +20,6 @@ import {
   InheritanceIndicator,
 } from '@/components/ui/inheritance-indicator';
 import { Label } from '@/components/ui/label';
-import {
-  Select,
-  SelectContent,
-  SelectItem,
-  SelectTrigger,
-  SelectValue,
-} from '@/components/ui/select';
 import { Separator } from '@/components/ui/separator';
 import { useFullAgentFormContext } from '@/contexts/full-agent-form';
 import { useRuntimeConfig } from '@/contexts/runtime-config';


agents-manage-ui/src/components/agent/sidepane/metadata/metadata-editor.tsx
```diff
@@ -388,36 +279,28 @@ export function MetadataEditor() {
...
+                    project?.stopWhen?.transferCountIs
+                  )}
+                  size="sm"
+                />
+              </>
+            }
+            name="stopWhen.transferCountIs"
             type="number"
-            min="1"
-            max="100"
```
dimaMachina, line 436: I removed because we have zod validation for min/max

agents-manage-ui/src/lib/actions/agent-full.ts
```diff
@@ -198,7 +162,7 @@ export async function updateFullAgentAction(
   projectId: string,
   agentId: string,
   agentData: FullAgentDefinition
-): Promise<ActionResult<FullAgentDefinition>> {
+): Promise<ActionResult<FullAgentResponse>> {
```
dimaMachina: mistake, should be response and not definition

agents-manage-ui/src/lib/actions/agent-full.ts
```diff
@@ -120,50 +120,14 @@ export async function updateAgentAction(
...
-
 /**
  * Get a full agent by ID
  */
 async function $getFullAgentAction(
   tenantId: string,
   projectId: string,
   agentId: string
-): Promise<ActionResult<FullAgentDefinition>> {
+): Promise<ActionResult<FullAgentResponse>> {
```
dimaMachina, line 108: mistake, should be response and not definition

agents-manage-ui/src/lib/types/agent-full.ts
```diff
@@ -38,10 +90,9 @@ export interface Agent {
 }
 
 // API Response Types
-export type CreateFullAgentResponse = SingleResponse<FullAgentDefinition>;
 export type CreateAgentResponse = SingleResponse<AgentApiInsert>;
-export type GetAgentResponse = SingleResponse<FullAgentDefinition>;
-export type UpdateFullAgentResponse = SingleResponse<FullAgentDefinition>;
+export type GetAgentResponse = SingleResponse<FullAgentResponse>;
+export type UpdateFullAgentResponse = SingleResponse<FullAgentResponse>;
```
dimaMachina, line 32: mistake, should be response

agents-manage-ui/src/components/agent/sidepane/metadata/context-config.tsx
```diff
@@ -1,54 +1,66 @@
...
+      callback({ values }) {
+        // Generate suggestions from context config
+        const variables = getContextSuggestions({
+          headersSchema: tryJsonParse(values.contextConfig?.contextVariables),
+          contextVariables: tryJsonParse(values.contextConfig?.headersSchema),
+        });
+        setVariableSuggestions(variables);
+      },
+    });
+  }, []);
```
dimaMachina, line 41: moved it here from https://github.com/inkeep/agents/blob/c120e78a700a56229c8aa51239895ae3e55cd6c7/agents-manage-ui/src/app/%5BtenantId%5D/projects/%5BprojectId%5D/agents/%5BagentId%5D/layout.tsx#L33-L43

agents-manage-ui/src/components/agent/sidepane/metadata/context-config.tsx
```diff
@@ -1,54 +1,66 @@
...
-  const updateContextConfig = (field: keyof ContextConfig, value: string) => {
-    const updatedContextConfig = {
-      ...contextConfig,
-      [field]: value,
-    };
-    updateMetadata('contextConfig', updatedContextConfig);
-  };
+  // biome-ignore lint/correctness/useExhaustiveDependencies: -- only on mount
+  useEffect(() => {
+    // make sure to unsubscribe;
```
sarah-inkeep, line 22: do we need to unsubscribe when this is unmounted?

dimaMachina:
> * I noticed that there are a few fields Im not able to remove after I've set the values:
>   context config, headers json, max transfers, provider options on the base model
>   example [loom](https://www.loom.com/share/4407cf65f52b49508e723b44bd36a7a7)
>   Could you go through and double check all of the fields are able to be removed, I may have missed some in my testing.
> * After I hit save "try it" is still disabled and I still see the `Unsaved changes` modal
> * Editing the `Provider options` field for the base model of an agent does not re-enable the "save" button
> * When I try updating a field like the agent description using the copilot the field is updated but it's not reflected in the UI, the UI shows the old description (it does work for the subagent fields so I think it's like something with the refactor?)
> * Can you take a look at the comments Claude left I think some may be accurate.

---

> I noticed that there are a few fields Im not able to remove after I've set the values:

I can confirm unable removing of context config and headers values.
You cannot delete max transfer, it's unset to 10
provider options on the base model can be removed now

> After I hit save "try it" is still disabled and I still see the Unsaved changes modal

was fixed by reseting form with response from db

> Editing the Provider options field for the base model of an agent does not re-enable the "save" button

added `form.setValue('models.summarizer.providerOptions', value, { shouldDirty: true });` everywhere in metadata-editor

> When I try updating a field like the agent description using the copilot

need help with it

> do we need to unsubscribe when this is unmounted?

yes to follow best practices when component is unmount - stop listening

> Can you take a look at the comments Claude left I think some may be accurate.

yep, applied changes

agents-manage-ui/src/components/agent/form/validation.ts
```diff
@@ -0,0 +1,120 @@
...
+const ModelsBaseSchema = ModelsSchema.base.unwrap();
+const ModelsStructuredOutputSchema = ModelsSchema.structuredOutput.unwrap();
+const ModelsSummarizerSchema = ModelsSchema.summarizer.unwrap();
+
+const StringToJsonSchema = z
+  .string()
+  .trim()
+  .transform((value, ctx) => (value === '' ? undefined : transformToJson(value, ctx)))
+  .refine((v) => v !== null, 'Cannot be null')
+  .optional();
```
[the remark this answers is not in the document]
dimaMachina: `v !== null || v === null` <- wtf is this @claude

PR #1801, opened 2026-02-07
use builtin vitest file snapshots for OpenAPI json file

dimaMachina:
image

### DIFF
--- a/agents-api/vitest.config.ts
+++ b/agents-api/vitest.config.ts
@@ -18,7 +18,12 @@ export default defineProject({
     // Enable parallelism with in-memory databases - each worker gets isolated database
     fileParallelism: true,
     isolate: true, // Ensure test isolation to prevent state leakage
-    maxWorkers: 10, // Increase for GitHub Actions runners (have more cores)
+    poolOptions: {
+      threads: {
+        maxThreads: 10, // Increase for GitHub Actions runners (have more cores)
+        minThreads: 4,
+      },
+    },
     env: {
       ENVIRONMENT: 'test',
       ANTHROPIC_API_KEY: 'test-api-key',
--- a/agents-api/package.json
+++ b/agents-api/package.json
@@ -99,12 +99,12 @@
     "@types/jmespath": "^0.15.2",
     "@types/node": "^20.11.24",
     "@types/pg": "^8.15.6",
-    "@vitest/coverage-v8": "^4.0.18",
+    "@vitest/coverage-v8": "^3.2.4",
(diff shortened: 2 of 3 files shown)


agents-api/vitest.config.ts
```diff
@@ -18,12 +18,7 @@ export default defineProject({
     // Enable parallelism with in-memory databases - each worker gets isolated database
     fileParallelism: true,
     isolate: true, // Ensure test isolation to prevent state leakage
-    poolOptions: {
-      threads: {
-        maxThreads: 10, // Increase for GitHub Actions runners (have more cores)
-        minThreads: 4,
```
dimaMachina, line 24: looks like this option was removed in vitest 4

dimaMachina:
@claude verify and wdyt

agents-api/src/domains/run/agents/Agent.ts
```diff
@@ -39,9 +39,9 @@ import {
   type ToolSet,
   tool,
 } from 'ai';
-import manageDbPool from 'src/data/db/manageDbPool';
-import { env } from 'src/env';
+import manageDbPool from '../../../data/db/manageDbPool';
```
amikofalvy, line 42: @dimaMachina Why are we using relative references instead of path alias?

dimaMachina: we don't use path aliases in this package, it was silently working with tsdown I guess, I updated vitest to v4 and it complained about these paths, but I did rollback of vitest v4 due some typecheck problems

agents-api/src/__tests__/openapi.test.ts
```diff
@@ -251,119 +124,19 @@ describe('OpenAPI Specification - Integration Tests (Unified agents-api)', () =>
...
-        lines.push('  │                                                                 │');
-        lines.push('  │  pnpm --filter @inkeep/agents-api openapi:update-snapshot       │');
-        lines.push('  └─────────────────────────────────────────────────────────────────┘');
-        lines.push('');
-        lines.push('═'.repeat(70));
-        lines.push('');
-
-        throw new Error(lines.join('\n'));
-      }
+      const openApiJsonSnapshot = `${JSON.stringify(sortObjectKeys(cachedSpec), null, 2)}\n`;
```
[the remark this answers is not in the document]
amikofalvy: @dimaMachina Did you consider this feedback? I remember building in the normalization when I first built the snapshot test. Is it not required anymore with vitest or something?

PR #1836, opened 2026-02-09
Normalize JSON schemas for OpenAI structured output compatibility

tim-inkeep:
Azure OpenAI's structured output API requires ALL properties to be in the required array, even nullable/optional ones. When data components with optional properties (like fact: z.string().nullable()) were converted to JSON schemas, they failed with:

`Invalid schema for response_format 'response': 'required' is required to be supplied and to be an array including every key in properties. Missing 'fact'.`

Solution:
Added SchemaProcessor.normalizeForStructuredOutput() that recursively ensures all properties in JSON schemas are marked as required before conversion to Zod schemas via z.fromJSONSchema().                                                     

Changes:
  - SchemaProcessor.ts: Added normalization method that makes all schema properties required
  - Agent.ts: Apply normalization to data component schemas
  - artifact-component-schema.ts: Apply normalization to artifact component schemas
  - Works for all providers (OpenAI/Azure require it, Anthropic accepts it)

### DIFF
--- a/.changeset/large-words-dream.md
+++ b/.changeset/large-words-dream.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+fix max height for card content on traces conversation page
--- a/agents-manage-ui/src/app/[tenantId]/projects/[projectId]/traces/conversations/[conversationId]/page.tsx
+++ b/agents-manage-ui/src/app/[tenantId]/projects/[projectId]/traces/conversations/[conversationId]/page.tsx
@@ -259,7 +259,7 @@ export default function ConversationDetail({
 
           return (
             <Card
-              className="shadow-none bg-background"
+              className="shadow-none bg-background max-h-[280px] flex flex-col"
               title={
                 conversation.conversationStartTime
                   ? `Start: ${formatDateTime(conversation.conversationStartTime, { local: true })}${showEndTime && conversation.conversationEndTime ? `\nEnd: ${formatDateTime(conversation.conversationEndTime, { local: true })}` : ''}`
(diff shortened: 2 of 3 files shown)


agents-api/src/domains/manage/routes/projectFull.ts
```diff
@@ -162,6 +164,35 @@ app.openapi(
         projectData: validatedProjectData,
       });
 
+      // 4. Sync to SpiceDB: link project to org and grant creator admin role
+      // CRITICAL: If this fails, the project exists but the creator can't access it
```
amikofalvy: If this comment is true, then the application should try to recover from this by wrapping both operations in a single "Transaction". I put it in quotes because I know we can't use the postgres built in transactions across multiple types of databases. 

`try { operation } catch () { tx.rollback(); }`

agents-api/src/domains/run/utils/SchemaProcessor.ts
```diff
@@ -223,6 +224,59 @@ export class SchemaProcessor {
...
 
+  /**
+   * Normalize JSON schema for structured output compatibility
+   * Makes all properties required - this ensures compatibility across all LLM providers
+   * (OpenAI/Azure require it, Anthropic accepts it)
+   *
+   * @param schema - JSON schema to normalize (supports nested objects, arrays, unions)
+   * @returns Normalized schema with all properties marked as required
+   */
+  static normalizeForStructuredOutput(
```
amikofalvy: I find that `normalizeForStructuredOutput` name hides the behavior of the function which is to make all properties required. I suggest renaming the function for clarity. `makeAllPropertiesRequired(` which leaves no shadow of doubt. Then the jsdoc can be removed and the function signature is enough to describe the entire function behavior.

amikofalvy:
this PR appears to combine multiple separate issues in a single batch:
1. function that makes all properties required in a json schema (for compatibility purposes with certain models)
2. SpiceDB Sync

Can you explain why these are both in here? Is the spice db change related to the json schema change in some way I don't know about?

tim-inkeep:
> this PR appears to combine multiple separate issues in a single batch:
> 
> 1. function that makes all properties required in a json schema (for compatibility purposes with certain models)
> 2. SpiceDB Sync
> 
> Can you explain why these are both in here? Is the spice db change related to the json schema change in some way I don't know about?

@amikofalvy i noticed both of these issues at the same time when i was testing an agent for the first time and both were causing me issues with the agent i tested so both were fixed at the same time

amikofalvy (approved):

amikofalvy:
Closing in favor of the two split PRs

PR #1838, opened 2026-02-09
fix: enhance tool approval handling, so agents are conversationally aware of approvals

anubra266:


agents-api/src/domains/run/utils/tool-result.ts
```diff
@@ -0,0 +1,24 @@
...
+  toolCallId: string;
+  reason?: string;
+}
+
+export function isToolResultDenied(result: unknown): result is DeniedToolResult {
+  return (
+    !!result &&
+    typeof result === 'object' &&
+    INKEEP_TOOL_DENIED_KEY in (result as object) &&
+    (result as Record<string, unknown>)[INKEEP_TOOL_DENIED_KEY] === true
```
dimaMachina: should not have any type errors since you have `typeof result === 'object'` check

```suggestion
    INKEEP_TOOL_DENIED_KEY in result &&
    result[INKEEP_TOOL_DENIED_KEY] === true
```

.changeset/slow-loops-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-api": patch
+---
+
+enhance tool approval handling, so agents are conversationally aware of approvals
```
amikofalvy, line 5: Needs another changeset file for the distinct changes to the agents core package

agents-api/templates/v1/prompt/system-prompt.xml
```diff
@@ -51,6 +51,12 @@
       - NEVER mention delegation occurred: just present the results
       - If delegation returns artifacts, reference them as if you created them
 
+      🚨 TOOL APPROVAL DENIAL - CRITICAL:
+      - Some tools require user approval before running. The user can deny approval.
+      - When a tool result says the user declined to run the tool (e.g. "The user declined to run this tool" or indicates denial), the user intentionally chose not to run it.
+      - Acknowledge that the user chose not to run the tool. Do NOT suggest technical failures, service unavailability, or errors.
+      - Offer to try again when they are ready or to help with something else.
```
amikofalvy: I think the default prompt for tool rejection should include asking for additional user input on what to expect. Tool rejection should have the optional reason for the rejection. Simple "reject" button will not have a reason, but follow up user messages can indicate a reason.

agents-api/src/domains/run/agents/Agent.ts
```diff
@@ -2373,6 +2363,20 @@ export class Agent {
   private formatToolResult(toolName: string, args: any, result: any, toolCallId: string): string {
     const input = args ? JSON.stringify(args, null, 2) : 'No input';
 
+    if (isToolResultDenied(result)) {
+      const reason = result.reason ?? 'User denied approval';
+      const output = `The user declined to run this tool. Reason: ${reason}`;
+      return `## Tool: ${toolName}
+
```
amikofalvy: Nit: The multiline template literal here breaks out of the surrounding indentation, which is a well-known readability issue with template literals in TypeScript/JavaScript — and something we have scattered throughout the codebase.

We could clean this up with an array join pattern:

```suggestion
    return [
      `## Tool: ${toolName}`,
      '',
      `### 🔧 TOOL_CALL_ID: ${toolCallId}`,
      '',
      `### Input`,
      `${input}`,
      '',
      `### Output`,
      `${output}`,
    ].join('\n');
```

This keeps everything aligned with the surrounding code. Not blocking on this — just flagging it as a pattern we could start adopting to gradually clean up the codebase.

agents-api/templates/v1/prompt/system-prompt.xml
amikofalvy: This can be done in a future commit

amikofalvy: nvm, skip this scope of work as per our slack huddle.

PR #1864, opened 2026-02-10
fixed formatting on docs files

amikofalvy:


amikofalvy:
@claude The docs should never have landed on main branch without being run through the formatter. Figure out which commit caused this issue in the first place and determine how we can stop it from occurring again.

PR #1875, opened 2026-02-10
feat: add anonymous user support with JWE tokens and SpiceDB authorization

amikofalvy:
## Summary

- **Anonymous users table + data access**: New `anonymous_users` runtime DB table with composite primary key (`tenantId`, `projectId`, `id`), Drizzle migration, and CRUD data-access functions.
- **JWE token authentication**: `jose`-based encrypted token generation/verification (`dir` + `A256GCM`) for anonymous user identity, with `isAnonymousToken` discriminator and dedicated Hono middleware (`requireAnonymousAuth` / `optionalAnonymousAuth`).
- **SpiceDB authorization**: Added `anonymous_user` and `conversation` definitions to `schema.zed` with participant/project relations and view/delete permissions. Updated `types.ts` constants to match.
- **Unit tests**: 13 new tests covering JWE token round-trips, anonymous user data access, and conversation listing with `anonymousUserId` filtering.

## Test plan

- [x] `anonymous-token.test.ts` — token generation, verification, tamper detection, expiry, `isAnonymousToken`
- [x] `anonymousUsers.test.ts` — create, get, metadata handling
- [x] `conversations.test.ts` — existing tests still pass with new `anonymousUserId` column
- [x] OpenAPI snapshot updated
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm format`, `pnpm build` all pass
- [ ] Manual smoke test: hit anonymous chat endpoint, verify token issuance, list conversations with returned token


nick-inkeep:
**Suggestions // Discussion Topics below**

---

## 1. Public client identity + agent binding (App model)

**Proposal**: Add a runtime DB `apps` entity with `appType='public_client'` and `bindingType='agent'`, where `clientId` is the only embed-time identifier and the binding is inherent (`tenantId`, `projectId`, `agentId` live on the row and are immutable for this app type).

```ts
type App = {
  id: string;
  clientId: string; // globally unique, public
  tenantId: string;
  projectId: string;
  agentId: string; // inherent binding, not a "default"
  appType: 'public_client';
  bindingType: 'agent';
  anonymousUsersEnabled: boolean;
  allowedOrigins: string[]; // exact origins, canonical form
  allowLocalhost: boolean;
  disabledAt?: string | null;
  createdAt: string;
  updatedAt: string;
};
```

**Reasoning**: The public widget should only know `clientId` + bearer token. Server-side resolution of `tenantId`, `projectId`, `agentId` from `clientId` prevents leaking internal identifiers and removes the need for `x-inkeep-tenant-id`, `x-inkeep-project-id`, `x-inkeep-agent-id` headers from anonymous clients. Using `agentId` (required) rather than `defaultAgentId` (optional) matches the invariant that a public client app only ever operates on exactly one agent.

**Current**: No `App` entity exists. Anonymous bootstrap and anonymous JWE auth both rely on the client sending `x-inkeep-tenant-id`, `x-inkeep-project-id`, and `x-inkeep-agent-id` as request headers. This exposes internal identifiers to browser JS and allows anyone with knowledge of valid IDs to mint anonymous tokens for arbitrary projects/agents.

```369:386:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/middleware/runAuth.ts
if (!reqData.tenantId || !reqData.projectId || !reqData.agentId) {
    throw new HTTPException(400, {
      message:
        'x-inkeep-tenant-id, x-inkeep-project-id, and x-inkeep-agent-id headers are required for anonymous access',
    });
  }
```

**Alternatives considered**:
- **Project-scoped app + `defaultAgentId`**: App is project-bound and chooses an agent when unspecified. **Why not**: implies the app may operate on other agents; the stated invariant is "only works for that agent."
- **Separate `app_bindings` table now**: 1 app to many bindings (installation model). **Why not**: the widget would need a binding selector beyond `clientId`, conflicting with "client sees only `clientId` + bearer token."

---

## 2. Anonymous bootstrap flow (no internal IDs from client)

**Proposal**: Replace the `x-inkeep-anonymous` + internal-header bootstrap with a single explicit bootstrap endpoint that accepts only `clientId`, resolves the App server-side, and returns the anonymous token in the response body.

```ts
// POST /run/api/anonymous/bootstrap
type BootstrapRequest = { clientId: string };

type BootstrapResponse = {
  anonymousToken: string; // JWE now; potentially opaque later
};
```

**Reasoning**: The server derives `tenantId`, `projectId`, and `agentId` from the `clientId` lookup, so the client never needs to know or transmit them. This prevents token minting for arbitrary projects/agents if internal IDs are leaked or guessable.

**Current**: `tryAnonymousNewAuth` creates an anonymous auth context from `x-inkeep-anonymous: true` combined with `x-inkeep-tenant-id`, `x-inkeep-project-id`, and `x-inkeep-agent-id` headers. No validation that the caller is authorized for that tenant/project/agent combination.

**Alternatives considered**:
- **Keep header-based bootstrap but add an "anonymous enabled" gate**: still requires the client to send `tenantId`, `projectId`, `agentId`. **Why not**: still exposes internal identifiers to the browser.
- **OAuth consent-only flow (no login)**: user clicks "Allow" in a popup. **Why not (for this scenario)**: public docs widget visitors cannot be shown an OAuth consent screen.

---

## 3. Anonymous token delivery (response body, not response header)

**Proposal**: Return the anonymous token in the **JSON response body**. Do not require the client to read a custom response header.

```ts
return c.json({ anonymousToken });
```

**Reasoning**: Cross-origin browser JS cannot read non-safelisted response headers unless the server sets `Access-Control-Expose-Headers`. The current `/run` CORS config does not expose `x-anonymous-token`, so browser clients cannot read it. Returning tokens in the body avoids this class of issue entirely and matches OAuth token endpoint conventions (RFC 6749 Section 5.1).

**Current**: The token is set via `c.header('x-anonymous-token', anonUser.token)` in both `chat.ts` and `chatDataStream.ts`, but `runCorsConfig.exposeHeaders` only includes `Content-Length`.

```240:243:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/domains/run/routes/chat.ts
const anonUser = await resolveAnonymousUser(executionContext);
if (anonUser?.isNew && anonUser.token) {
  c.header('x-anonymous-token', anonUser.token);
}
```

**Alternatives considered**:
- **Header only (`x-anonymous-token`)**: requires adding it to `exposeHeaders` and fixing CORS origin echoing. **Why not**: fragile in cross-origin browser scenarios; easy to silently break.
- **Both header + body**: body as canonical, header for convenience. **Why not (phase 1)**: adds two delivery mechanisms while CORS configuration is still being reworked.

---

## 4. CORS + per-App origin enforcement

**Proposal**: Replace the global permissive `/run` CORS config with dynamic, per-App origin enforcement:
- Echo only validated origins from `apps.allowedOrigins` (exact match).
- Add `Vary: Origin` to responses.
- Reject `Origin: null`.
- Do not use `credentials: true` for the widget flow (no cookies).
- Enforce origin in both CORS middleware (browser correctness) and auth middleware (defense-in-depth against non-browser callers).

```ts
function isAllowedOriginForApp(origin: string, app: App): boolean {
  if (origin === 'null') return false;

  const effective = canonicalizeOrigin(origin);
  if (app.allowedOrigins.includes(effective)) return true;

  if (app.allowLocalhost && isLocalhostOrigin(effective)) return true;

  return false;
}
```

**Reasoning**: The widget is cross-origin by definition (`docs.customer.com` to `api.inkeep.com`). Per-client origin allowlisting is the primary abuse boundary for public clients. `allowLocalhost` is a separate, explicit flag so local development access is auditable and can trigger different rate-limit policies.

**Current**: `runCorsConfig` sets `origin: '*'` with `credentials: true`. Browsers reject this combination. There is no per-App origin enforcement; all origins are accepted for `/run`.

```119:126:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/middleware/cors.ts
export const runCorsConfig: CorsOptions = {
  origin: '*',
  allowMethods: ['GET', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'PATCH'],
  allowHeaders: ['*'],
  exposeHeaders: ['Content-Length'],
  maxAge: 86400,
  credentials: true,
};
```

**Alternatives considered**:
- **Wildcard subdomains (`*.example.com`)**: dynamic suffix matching with origin echoing. **Why not**: explicitly deferred; exact allowlist only for phase 1 to avoid regex/suffix-matching pitfalls.
- **Keep global permissive `/run` CORS**: allow all origins. **Why not**: incompatible with per-client threat model; enables unrestricted use from arbitrary origins.
- **Allow `Origin: null`**: dev convenience. **Why not**: `null` origin is a known security footgun (sandboxed iframes, data URIs produce `null`; OWASP flags this explicitly).

---

## 5. Anonymous token claim standardization

**Proposal**: Standardize anonymous token claims to include `tokenUse`, `aud`, and `client_id`, aligning with existing Slack token patterns and RFC 9068 (JWT Access Token Profile).

```ts
type AnonymousAccessTokenClaims = {
  iss: 'inkeep-auth';
  aud: 'inkeep-api';
  tokenUse: 'anonymousUser';
  client_id: string; // app.clientId
  sub: string; // anonymousUserId
  iat: number;
  exp: number;
  jti: string;
};
```

**Reasoning**: `tokenUse` provides an explicit discriminator for how to interpret `sub` and which auth rules apply (the same pattern used for `slackUser` tokens). `client_id` is a required claim per RFC 9068 and enables attribution/policy enforcement. `aud` prevents cross-service token confusion. Including these claims keeps future STS/token-exchange and MCP/OAuth flows aligned. Keeping `tenantId`/`projectId` *out* of the token is intentional: since the App binding is immutable, these are derived server-side from `client_id`, which also makes a future JWE-to-opaque swap low-risk.

**Current**: Anonymous JWE payload contains `tenantId` and `projectId` as custom claims and uses `iss='inkeep-anonymous'`. It does not include `aud`, `tokenUse`, `client_id`, or `jti`.

```44:53:/Users/edwingomezcuellar/InkeepDev/agents/packages/agents-core/src/utils/anonymous-token.ts
const token = await new EncryptJWT({
    tenantId: params.tenantId,
    projectId: params.projectId,
  })
    .setProtectedHeader({ alg: 'dir', enc: 'A256GCM' })
    .setSubject(params.anonymousUserId)
    .setIssuer(ISSUER)
    .setIssuedAt()
    .setExpirationTime(DEFAULT_EXPIRY)
    .encrypt(key);
```

**Alternatives considered**:
- **Embed `tenantId`, `projectId`, `agentId` in the token (current approach)**: token carries binding context. **Why not**: redundant when binding is immutable and resolvable from `client_id`; also makes a swap to opaque tokens harder.
- **Switch to opaque tokens now**: random handle mapped to server-side session. **Why not (immediately)**: deferred as explicit open question; the system should be designed so swapping is low-risk, but does not need to ship opaque in phase 1.

---

## 6. Token format (JWE vs opaque) — explicit open question

**Proposal**: Keep JWE for phase 1 but design the system so swapping to opaque tokens later requires minimal changes (i.e., binding and ownership live in the DB + App model, not inside the token).

**Reasoning**: Opaque tokens simplify revocation, rotation, and abuse controls. JWE hides claim contents from browser JS but does not prevent token theft under XSS (it hides contents, not usability). The current architecture should not depend on being able to decrypt/read token contents on every request path.

**Current**: JWE is implemented and used.

**Alternatives considered**:
- **Opaque token now**: server-side session table. **Why not (immediately)**: requires designing session model (TTL, cleanup, token hashing at rest). Decision deferred.
- **JWS (signed, readable)**: simpler than JWE. **Why not**: exposes `anonymousUserId` and any internal claims to browser JS.

---

## 7. Conversation ownership enforcement (DB filter, not SpiceDB)

**Proposal**: For anonymous principals, enforce conversation ownership strictly via DB equality checks (`conversations.anonymous_user_id = $anonymousUserId`), not via per-conversation SpiceDB relationships. Apply this check on **every** endpoint that accepts a `conversationId` when the principal is anonymous.

```sql
SELECT * FROM conversations
WHERE tenant_id = $tenantId
  AND project_id = $projectId
  AND id = $conversationId
  AND anonymous_user_id = $anonymousUserId;
```

**Reasoning**: Writing a SpiceDB `participant` relationship for every anonymous conversation creates high-cardinality relationship graphs (potentially millions of relationships). The DB already stores `conversations.anonymousUserId`, making equality checks simple and fast. SpiceDB remains appropriate for broader admin-style access policies (e.g., "org admins can view all conversations").

**Current**: `writeAnonymousConversationRelationships` writes per-conversation `participant` relationships to SpiceDB. Conversation listing tries `lookupResources` from SpiceDB first and falls back to a DB query filtered by `anonymousUserId`.

```62:77:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/domains/run/utils/anonymous-user.ts
await writeRelationship({
  resourceType: SpiceDbResourceTypes.CONVERSATION,
  resourceId: conversationId,
  relation: 'participant',
  subjectType: SpiceDbResourceTypes.ANONYMOUS_USER,
  subjectId: anonymousUserId,
});
```

```69:76:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/domains/run/routes/conversations.ts
conversationIds = await lookupResources({
  resourceType: SpiceDbResourceTypes.CONVERSATION,
  permission: SpiceDbConversationPermissions.VIEW,
  subjectType: SpiceDbResourceTypes.ANONYMOUS_USER,
  subjectId: anonymousUserId,
});
```

**Alternatives considered**:
- **Model anonymous participation fully in SpiceDB**: write `participant` relationships for all conversations. **Why not**: high cardinality; the DB ownership column is already present and more efficient for this access pattern.
- **No ownership check (trust `conversationId` unguessability)**: rely on random IDs being hard to guess. **Why not**: IDOR risk; authorization must be enforced regardless of ID entropy.

---

## 8. Remove `x-inkeep-agent-id` requirement for anonymous auth

**Proposal**: Anonymous auth should not require the client to send `x-inkeep-agent-id`. The agent is determined from the App binding resolved by `clientId` (or by the token's `client_id` claim on subsequent requests).

```ts
// in auth middleware (pseudo)
const app = await resolveAppByClientId(token.client_id);
const { tenantId, projectId, agentId } = app; // inherent binding
```

**Reasoning**: The widget should not know `agentId`. Requiring `x-inkeep-agent-id` today leaks an internal identifier and is redundant when the App is agent-bound.

**Current**: `tryAnonymousJweAuth` reads `agentId` from `reqData.agentId` (the `x-inkeep-agent-id` header) and throws if it is missing.

```340:348:/Users/edwingomezcuellar/InkeepDev/agents/agents-api/src/middleware/runAuth.ts
const agentId = reqData.agentId;

if (!agentId) {
  logger.warn({}, 'Anonymous JWE auth missing x-inkeep-agent-id header');
  throw new HTTPException(400, {
    message: 'x-inkeep-agent-id header is required for anonymous authentication',
  });
}
```

**Alternatives considered**:
- **Keep requiring `x-inkeep-agent-id`**: client controls agent selection. **Why not**: contradicts single-agent binding and leaks an internal identifier.
- **Embed `agentId` inside the token**: no header required; token carries it. **Why not (with agent-bound App)**: redundant; better derived server-side from App binding to keep tokens minimal and future opaque-compatible.

---

## 9. API key to App migration (backfill)

**Proposal**: Migrate immediately by mapping existing hashed API keys into `app_secret_keys` records attached to new `apps` rows (one App per unique `(tenantId, projectId, agentId)` in `api_keys`). Validation behavior stays identical: O(1) lookup by `publicId` + `scrypt` hash compare via `timingSafeEqual`.

```ts
type AppSecretKey = {
  id: string;
  appId: string; // FK to apps
  publicId: string; // from api_keys.public_id
  keyHash: string; // from api_keys.key_hash (scrypt, unchanged)
  keyPrefix: string;
  expiresAt?: string | null;
  lastUsedAt?: string | null;
  createdAt: string;
  updatedAt: string;
};
```

**Reasoning**: Existing API keys are one-way hashed (`scrypt`), so plaintext secrets cannot be recovered. Migration must be a mapping/backfill, not a re-issuance. Unifying secret keys under `apps` aligns the credential model so that `sk_...` and `clientId` resolve through the same App entity and the same `AuthContext` pipeline.

**Current**: API keys are stored hashed in `api_keys` with `publicId` for O(1) lookup. Each key is scoped to `(tenantId, projectId, agentId)`. There is no `App` entity.

```148:171:/Users/edwingomezcuellar/InkeepDev/agents/packages/agents-core/src/db/runtime/runtime-schema.ts
export const apiKeys = pgTable(
  'api_keys',
  {
    ...projectScoped,
    agentId: varchar('agent_id', { length: 256 }).notNull(),
    publicId: varchar('public_id', { length: 256 }).notNull().unique(),
    keyHash: varchar('key_hash', { length: 256 }).notNull(),
    keyPrefix: varchar('key_prefix', { length: 256 }).notNull(),
    // ...
  }
);
```

**Alternatives considered**:
- **Dual-stack indefinitely** (keep `api_keys` table + add `apps` separately, two auth paths). **Why not**: maintains two mental models and delays convergence of auth pipeline.
- **Re-issue new secrets for all API keys**: generate new keys and distribute to customers. **Why not**: impossible due to one-way hashing; plaintext secrets are not stored.

---

## 10. Exact-origin matching + canonicalization

**Proposal**: Canonicalize `Origin` consistently using `scheme + host + port` (with default ports omitted), aligning with browser origin semantics (MDN/RFC 6454). Store `apps.allowedOrigins` in the same canonical form and do exact string comparison.

```ts
function canonicalizeOrigin(origin: string): string {
  const u = new URL(origin);
  const scheme = u.protocol.toLowerCase();
  const host = u.hostname.toLowerCase();

  const port = u.port || (scheme === 'https:' ? '443' : scheme === 'http:' ? '80' : '');
  const isDefault =
    (scheme === 'https:' && port === '443') || (scheme === 'http:' && port === '80');

  return isDefault ? `${scheme}//${host}` : `${scheme}//${host}:${port}`;
}
```

**Reasoning**: Browsers define origin equality as (scheme, hostname, port). Default ports are implied by scheme, so `https://a.com` and `https://a.com:443` are the same origin. Canonicalization prevents accidental mismatches and reduces operator confusion when configuring allowed origins.

**Current**: No per-App origin enforcement for `/run`. The global `isOriginAllowed` function in `cors.ts` checks same-domain and localhost, but does not do per-App allowlisting or canonical comparison.

**Alternatives considered**:
- **String compare without canonicalization**: store raw origin strings. **Why not**: `https://a.com` vs `https://a.com:443` mismatch causes silent failures.
- **Wildcard subdomain matching (`*.example.com`)**: dynamic suffix match. **Why not**: explicitly deferred; exact allowlist only for phase 1 to avoid regex/suffix-matching bypass risks (OWASP flags this explicitly).

PR #1348, opened 2025-12-19
Agent Skills

dimaMachina:

PR #1895, opened 2026-02-10
make `<ExpandablePromptEditor />` work on markdown files without `add variables` action button

dimaMachina:
part of https://github.com/inkeep/agents/pull/1348 where `ExpandablePromptEditor` can be simple markdown editor


dimaMachina:

PR #1909, opened 2026-02-10
Bugfix/artifact saving tool ids

tim-inkeep:


omar-inkeep:
I addressed the suggestions for the spicedb-related changes. I think we can merge this

PR #1910, opened 2026-02-10
warn if user try to attach edge for already connected MCP, remove `0` badge in MCP nodes, in favour of `No tools`

dimaMachina:
https://github.com/user-attachments/assets/fa476489-cf9a-4d3a-bf99-e205192292f2


dimaMachina:
@claude --verify and add a changeset

PR #1921, opened 2026-02-11
adjust editors styles in full screen mode

dimaMachina:


dimaMachina:
@claude --review and add a changeset

PR #1924, opened 2026-02-11
Disable automatic layout in Monaco editor

dimaMachina:
Disable automatic layout to prevent sidebar flickering.


dimaMachina:
@claude add a changeset

PR #1925, opened 2026-02-11
Disable automatic layout in diff editor too

dimaMachina:
similar to https://github.com/inkeep/agents/pull/1924


dimaMachina:
@claude add a changeset

dimaMachina:
found better fix https://github.com/inkeep/agents/pull/1960 but needs patch monaco-editor

PR #1927, opened 2026-02-11
Image support: blob storage and media proxy

mike-inkeep:
**Summary**  
Persist chat images in blob storage instead of inline base64 and serve them via an authenticated media proxy. Supports S3-compatible (MinIO/AWS) and Vercel Blob.

**Changes**

- **Blob storage**
  - New `BlobStorageProvider` abstraction with **S3** and **Vercel Blob** implementations.
  - Env: `BLOB_STORAGE_PROVIDER`, `BLOB_*` (S3 endpoint, bucket, creds; optional `BLOB_READ_WRITE_TOKEN` for Vercel). Documented in `.env.example`.
  - Docker Compose: MinIO service + `minio-setup` that creates the `inkeep-agents-media` bucket for local dev.

- **Message content and upload**
  - User messages with image parts (inline base64 or external URLs) are run through `buildPersistedMessageContent`: images are uploaded to blob storage and stored as `blob://` URIs; message content is persisted as `{ text, parts }` with file parts pointing at those URIs (no base64 in the DB). Wired into both chat and chat streaming handlers.

- **Security (image-security)**
  - External image URLs: only HTTP/HTTPS, no credentials, allowed ports; DNS resolution blocks private/reserved IPs (SSRF); redirects re-validated; content-type and magic-byte checks; 10 MB limit; allowlist of image MIME types. Inline images: size limit and type validation. Failed or blocked uploads result in text-only persisted content.

- **Media proxy and listing**
  - New route: `manage/tenants/:tenantId/projects/:projectId/conversations/:id/media/*` serves blobs by key (behind existing project “view” permission). When listing conversation messages, blob URIs in content are resolved to these proxy URLs via `resolveMessagesListBlobUris` so clients get stable, authenticated URLs.

- **Other**
  - Fix backspace escaping in `extractA2AMessageText` (regex `\b` → `\\b`).
  - Tests: image upload (external URL, private IP blocking, redirects, content-type, oversized inline), blob storage helpers, resolve-blob-uris, and `buildPersistedMessageContent`.
  - New deps: `@aws-sdk/client-s3`, `@aws-sdk/s3-request-presigner`, `@vercel/blob`, `file-type`, `ipaddr.js`. Workspace: zod override, because pulling in aws-sdk resulted in two incompatible versions of Zod which was causing `pnpm typecheck` to fill its heap and crash.

### DIFF
--- a/.changeset/fix-monaco-layout-flickering.md
+++ b/.changeset/fix-monaco-layout-flickering.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Fix Monaco editor automatic layout to prevent sidebar flickering
--- a/agents-manage-ui/src/components/editors/monaco-editor.tsx
+++ b/agents-manage-ui/src/components/editors/monaco-editor.tsx
@@ -98,7 +98,8 @@ export const MonacoEditor: FC<MonacoEditorProps> = ({
       language,
       // Shift the find widget left to avoid flicker when clicking its close button.
       extraEditorClassName: '[&_.find-widget]:!right-11',
-      automaticLayout: true,
+      // Keep disabled: automatic Monaco relayout causes sidebar flickering when hovering a collapsed sidebar.
+      automaticLayout: false,
       minimap: { enabled: false }, // disable the minimap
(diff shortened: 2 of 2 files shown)


mike-inkeep:
Breaking this up into smaller PRs

PR #1947, opened 2026-02-12
simplify SkillReference, remove `index` field, calculate based on array index

dimaMachina:


agents-cli/src/commands/pull-v3/components/__tests__/sub-agent-generator.test.ts
```diff
@@ -284,6 +284,31 @@ describe('Sub-Agent Generator', () => {
       expect(definition).not.toContain('legacyAgent2,');
     });
 
+    it('should generate skill references without index and keep index-based ordering', () => {
```
dimaMachina, line 287:
```suggestion
    it('should generate skill references and keep index-based ordering', () => {
```

PR #1960, opened 2026-02-12
Reduces unnecessary resize calls to avoid performance issues in monaco-editor

dimaMachina:


dimaMachina:

PR #1972, opened 2026-02-12
fix: preserve user image parts through agent transfers and delegations

mike-inkeep:
Images attached to user messages were silently dropped when a request
was transferred or delegated to another agent. Three root causes:

- Transfer loop rebuilt parts as text-only after the first iteration
- Delegation tool had no access to the caller's image parts
- blob:// URIs were passed directly to the AI SDK which cannot fetch them

Thread original non-text parts across transfer iterations, pass image
parts into delegation tools via closure, resolve blob:// URIs to the
media proxy before handing to the AI SDK, and refactor
buildPersistedMessageContent to return uploaded parts so the execution
pipeline carries lightweight blob URIs instead of raw base64.

### DIFF
--- a/agents-api/src/__tests__/run/agents/relationTools.test.ts
+++ b/agents-api/src/__tests__/run/agents/relationTools.test.ts
@@ -535,6 +535,7 @@ describe('Relationship Tools', () => {
           role: 'agent',
           content: {
             text: 'Test message',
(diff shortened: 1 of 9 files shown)


mike-inkeep:
Will be rebasing this on top of a new PR stack

PR #1986, opened 2026-02-13
feat: auto-login for Manage UI in local development

nick-inkeep:
## Problem

Every time a developer runs `pnpm dev` and opens the Manage UI at `localhost:3000`, they're greeted with a login screen. They must manually enter the credentials set by `pnpm db:auth:init` (or look them up in `.env`). This is pure friction — the credentials are already on the server, the session is local-only, and there's no security benefit to requiring manual entry in development.

**Additional problem (v2)**: The original implementation used a synthetic sign-in with `auth.handler()`, which required both `INKEEP_AGENTS_MANAGE_UI_USERNAME` and `INKEEP_AGENTS_MANAGE_UI_PASSWORD` env vars. If the user changed their password via the UI, the env var became stale and auto-login silently failed (401). This approach was fragile to any auth method evolution (SSO migration, MFA, OAuth).

## Solution

Automatically sign in the developer when `ENVIRONMENT=development` using Better Auth's `internalAdapter.createSession()` — a privileged server-side API that creates sessions without password verification.

1. **Server**: A `POST /api/auth/dev-session` endpoint that:
   - Reads only `INKEEP_AGENTS_MANAGE_UI_USERNAME` from env (no password needed)
   - Looks up the user via `ctx.internalAdapter.findUserByEmail(email)`
   - Creates a session via `ctx.internalAdapter.createSession(userId)` — triggers `databaseHooks.session.create.before` which sets `activeOrganizationId`
   - Signs the session token with HMAC-SHA-256 (matching Better Auth's internal format)
   - Returns `Set-Cookie` with the signed session token

2. **Client**: A `<DevAutoLoginProvider>` component that wraps the app in `layout.tsx`. On first load in development, it checks if the user is already authenticated. If not, it fires a single POST to `/api/auth/dev-session`, receives the session cookie, and reloads.

### Sequence

\`\`\`
Browser loads localhost:3000
  → DevAutoLoginProvider checks useAuthSession()
  → Not authenticated, not loading
  → POST /api/auth/dev-session (credentials: 'include')
    → Server reads INKEEP_AGENTS_MANAGE_UI_USERNAME from env
    → auth.$context → ctx.internalAdapter.findUserByEmail(email)
    → ctx.internalAdapter.createSession(userId)
      → triggers databaseHooks.session.create.before → sets activeOrganizationId
    → Signs token with HMAC-SHA-256 using ctx.secret
    → Returns Set-Cookie with signed session token
  → Browser receives session cookie
  → window.location.reload()
  → useAuthSession() now returns authenticated
  → Children render normally
\`\`\`

### Why internalAdapter (v2 evolution from synthetic sign-in)

The original implementation constructed a synthetic Request to auth.handler() with username+password. This was fragile:

| Scenario | Synthetic sign-in (v1) | internalAdapter (v2) |
|---|---|---|
| Password changed via UI | **Breaks** — env var stale | **Works** — no password needed |
| SSO/SAML migration | **Breaks** — password auth disabled | **Works** — operates below auth method layer |
| MFA enabled | **Breaks** — no TOTP to provide | **Works** — bypasses challenge flow |
| Google OAuth added | **Breaks** — password may not exist | **Works** — only needs email |
| Email changed | Both need env update | Both need env update |

\`internalAdapter.createSession()\` operates at the session layer — below all auth method logic. Maximally resilient to auth evolution.

### Safety boundaries

- **Development only**: Registered inside \`if (env.ENVIRONMENT === 'development')\` — route does not exist in production
- **Auth-gated**: Inside \`if (auth)\` — no route if auth is disabled
- **No credentials transmitted**: Client never sends credentials. Only email (in env) used server-side for lookup
- **Real sessions**: \`createSession()\` calls \`createWithHooks()\`, triggering org membership hooks
- **Client tree-shaken**: \`DevAutoLoginProvider\` returns children immediately in production
- **Graceful degradation**: Falls through to normal login page on any failure

### Core (v2)
- **\`agents-api/src/createApp.ts\`** — \`POST /api/auth/dev-session\` using \`internalAdapter.createSession()\` + HMAC-SHA-256 cookie signing
- **\`agents-manage-ui/src/components/providers/dev-auto-login-provider.tsx\`** — Client auto-login provider (unchanged from v1)

### Tests (rewritten for v2)
- **\`agents-api/src/__tests__/devSession.test.ts\`** — 8 tests: signature verification, cookie attributes, user lookup, error cases, env guards

### Docs & config (from v1)
- \`.env.example\`, \`.env.docker.example\`, troubleshooting, docker docs, contributing guide, create-agents README

## Test plan

- [x] 8/8 unit tests passing
- [x] Typecheck clean
- [x] Browser E2E: Chrome Incognito → localhost:3000 → auto-login → /default/projects
- [x] curl: POST /api/auth/dev-session → 200 + signed Set-Cookie
- [x] curl: GET /api/auth/get-session → valid session with activeOrganizationId
- [x] curl: Auth-protected route works with session cookie

### DIFF
--- a/agents-manage-ui/src/components/providers/dev-auto-login-provider.tsx
+++ b/agents-manage-ui/src/components/providers/dev-auto-login-provider.tsx
@@ -19,6 +19,12 @@ export function DevAutoLoginProvider({ children }: { children: React.ReactNode }
   useEffect(() => {
     if (process.env.NODE_ENV !== 'development') return;
 
+    // Skip auto-login under Cypress — it manages its own login flow via cy.login()
+    if ('Cypress' in window) {
+      setReady(true);
+      return;
+    }
+
     // Already authenticated — no auto-login needed
     if (isAuthenticated) {
       setReady(true);


omar-inkeep:
I haven't gotten a chance to pull and test it, but ideally this works for the quickstart too

nick-inkeep:
> I haven't gotten a chance to pull and test it, but ideally this works for the quickstart too

Yes, it does, was one of the test cases CC tested end to end (using the Claude browser extension). That was the main o.g. user story.

nick-inkeep:
Re: 💭 Consider item in review about `activeOrganizationId` — this finding is **incorrect**.

`internalAdapter.createSession()` **does** call `createWithHooks()`, which triggers `databaseHooks.session.create.before`. Evidence:

1. **Better Auth source** (`better-auth/dist/db/internal-adapter.mjs:128-145`):
```js
createSession: async (userId, ...) => {
    // ...builds session data...
    return await createWithHooks(data, "session", ...);
}
```

2. **Integration test** — `GET /api/auth/get-session` with the dev-session cookie returned:
```json
{
  "session": {
    "activeOrganizationId": "default"
  }
}
```

The `activeOrganizationId: "default"` confirms the hook fired and set org membership correctly. No changes needed.

nick-inkeep:
@claude --full-review

nick-inkeep:
@claude --full-review

nick-inkeep:
## Review findings addressed (9c28636)

All 4 findings from the latest full review have been accepted and fixed:

| Finding | Severity | Fix |
|---------|----------|-----|
| `authentication.mdx:98` — auto-login claim misleading in Docker context | Major | Clarified auto-login only works with `pnpm dev`, not Docker |
| `docker-local.mdx:99` — same Docker auto-login claim | Major | Removed auto-login claim, now says "sign in with admin credentials" |
| `troubleshooting.mdx:117` — password env var listed as required | Minor | Clarified only username is needed for auto-login; password only for `db:auth:init` |
| `devSession.test.ts:12` — mock type parameter consistency | Minor | Updated to `typeof import('@inkeep/agents-core')` matching codebase convention |

All review threads are resolved.

PR #1995, opened 2026-02-13
use widget for chat to edit

anubra266:
Closes PRD-5933


agents-manage-ui/src/components/agent/copilot/message-parts/tool-approval.tsx
```diff
@@ -34,7 +18,8 @@ interface EntityData {
 }
 
 interface ToolApprovalProps {
-  data: ToolCallApprovalData;
+  tool: ToolUIPart;
+  approve: (approved?: boolean) => Promise<void>;
   copilotAgentId?: string;
   copilotProjectId?: string;
   copilotTenantId?: string;
```
anubra266:
```suggestion

File: agents-manage-ui/src/components/agent/copilot/message-parts/tool-approval.tsx [hunk 2]
```diff
@@ -34,10 +18,8 @@ interface EntityData {
...
 interface ToolApprovalProps {
-  data: ToolCallApprovalData;
-  copilotAgentId?: string;
-  copilotProjectId?: string;
-  copilotTenantId?: string;
+  tool: ToolUIPart;
+  approve: (approved?: boolean) => Promise<void>;
   apiUrl?: string;
   cookieHeader?: string;
   copilotToken?: string;
```
anubra266:
```suggestion

sarah-inkeep (changes requested):
I noticed that there seems to be a difference in the tool names so the icon and diff aren't showing as expected:
For example:
Old:
Tool name shows as `Sub Agent` with the `<Bot />` icon and we see the text diff
Image

New:
Tool name shows as `Tool Sub Agent` so it falls back to the default icon and we don't see the diff between current prompt and the new one which we would expect since it is an `update`.
Image

sarah-inkeep (changes requested):
I'm seeing this error in the console when the tool approval ui shows:
`Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`
Here is a loom:
https://www.loom.com/share/922d7d2cb78948d184437fd1149ab593

I checked on the main branch and I don't see the error there.

PR #2050, opened 2026-02-17
dynamic env vars

robert-inkeep:
In the docker deployment, with a prebuilt nextjs app, need a way for the built nextjs app to use dynamic environment variables.


robert-inkeep:
@claude what are the performance implications? it reads from .env, so it's not really dynamic in the sense it can change whenever, only on initial server load

nick-inkeep:
Correct approach, merging [here](https://github.com/inkeep/agents/pull/2051).

PR #2056, opened 2026-02-17
feat: mock AI provider for run route testing without API keys

nick-inkeep:
## Summary

- Adds a built-in `mock` AI provider implementing `LanguageModelV2` for testing run routes without real API keys
- Registered in `ModelFactory` as `mock/{modelName}` alongside anthropic, openai, google, etc.
- 19 unit tests covering interface compliance, streaming, ModelFactory integration, and edge cases
- Docs updated in models.mdx

## Motivation

Run domain endpoints (`/v1/chat/completions`, `/api/chat`, A2A) require a real Anthropic/OpenAI API key to produce responses. This blocks:
- Local dev pipeline validation after `pnpm setup-dev`
- CI/CD integration tests for run routes
- Quick health checks on the run API without incurring LLM costs

## Approach

**Mock provider** (`packages/agents-core/src/utils/mock-provider.ts`):
- Implements `LanguageModelV2` from `@ai-sdk/provider`
- `doGenerate`: returns deterministic text with model name, message count, last user message, timestamp
- `doStream`: streams response line-by-line with small delays simulating real streaming
- Synthetic token usage (chars/4)
- No API keys required

**ModelFactory integration** (`packages/agents-core/src/utils/model-factory.ts`):
- Added `mock` to `BUILT_IN_PROVIDERS`
- `mock/default`, `mock/anything` all resolve via `createMockModel()`

## Test plan

19 unit tests in `packages/agents-core/src/__tests__/utils/mock-provider.test.ts`:
- [x] LanguageModelV2 interface compliance (specificationVersion, provider, modelId)
- [x] Any model name accepted under mock/ prefix
- [x] Structured mock response with all fields
- [x] Message count reflects prompt length
- [x] Synthetic token usage (inputTokens + outputTokens = totalTokens)
- [x] Long messages truncated to 200 chars
- [x] No user message → "(no user message)"
- [x] Custom model names in response
- [x] Streaming: multiple chunks with correct part types
- [x] Streaming: line-by-line deltas with newline separators
- [x] Streaming: conversation history message count
- [x] ModelFactory.parseModelString for mock/
- [x] ModelFactory.createModel produces working mock model
- [x] Mock model works without API keys
- [x] Mock model works with providerOptions present
- [x] Unsupported providers rejected
- [x] validateConfig passes for mock/default

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2063, opened 2026-02-17
feat: add --skip-provider CLI option and dashboard lap E2E test

nick-inkeep:
## Summary

Adds a `--skip-provider` option to `create-agents` CLI and extends the quickstart E2E test with a full dashboard lap — CLI scaffolding through browser login, navigation, and playground panel verification.

## Motivation

**User scenario 1: Zero-friction quickstart.** New users and CI pipelines shouldn't need a real API key to scaffold and explore a working project. `--skip-provider` lets you scaffold with the `mock/default` model (landed in #2056), so `pnpm create-agents my-project --skip-provider` produces a fully functional project end-to-end — including hitting the chat endpoint from the playground — without any provider credentials.

**User scenario 2: Full-journey regression coverage.** The existing E2E test stopped at "API returns project metadata." That left the dashboard login → project navigation → agent detail → playground flow completely uncovered. This PR extends coverage to catch regressions across the entire first-run experience a user would walk through.

### `--skip-provider` CLI option

Two entry points, same behavior:

1. **`--skip-provider` flag** — non-interactive, scaffolds with `{ base: { model: 'mock/default' } }` model config
2. **Interactive "Skip" option** — 5th choice in the provider select prompt, same result

`--skip-provider` takes precedence over any `--openai-key`/`--anthropic-key` flags passed alongside it (logs a warning rather than silently ignoring).

### Dashboard lap E2E

The test now continues after `inkeep push` succeeds:

1. Starts the manage-ui standalone server (`startDashboardServer()` — forks the Next.js standalone `server.js` from the built `@inkeep/agents-manage-ui` package)
2. Launches headless Chromium via Playwright
3. Logs in via email/password form (`admin@example.com`)
4. Navigates: projects list → project detail → agent detail → "Try it" button → playground panel
5. Asserts `#playground-pane` visible and `#inkeep-widget-root` initialized

On failure: captures full-page screenshot + page URL + HTML content for debugging.

## Architectural decisions

**Playwright over Cypress for E2E dashboard lap.** The existing Cypress test suite is for the manage-ui package itself. This test lives in `create-agents` and verifies the scaffolded project's dashboard works end-to-end. Playwright is lighter to install in CI (just chromium), doesn't need a dev server, and works well with the standalone Next.js output.

**Standalone server, not `next dev`.** The dashboard is started from the pre-built `.next/standalone/agents-manage-ui/server.js` via `node:child_process.fork()`. This is faster and more realistic — it tests the same artifact that ships to production.

**`localhost` vs `127.0.0.1` split.** The API server uses `127.0.0.1:3002` (avoids IPv6 resolution issues on Ubuntu CI), but the dashboard must use `localhost:3002` so auth cookies share the same domain between dashboard (`:3000`) and API (`:3002`). Both are `localhost`-scoped cookies.

**Auth init reliability.** CI applies migrations from the monorepo root before the test runs, so `setup-dev:cloud`'s internal migration step can fail and exit early — skipping auth init entirely. Fixed by running auth init as a separate step. Also added an idempotent signup-via-API fallback (`POST /api/auth/sign-up/email`) as a belt-and-suspenders measure.

**Selector strategy.** Card components use a link-overlay pattern that intercepts pointer events, so clicks use `a[href*="..."]` with `force: true` rather than text-based selectors. `waitForURL` uses `domcontentloaded` instead of `load` to avoid timeout from slow external resource loading.

## Known limitation

Message sending in the playground (typing + receiving a response via the chat widget) is not yet tested. The `InkeepEmbeddedChat` widget renders in a shadow DOM / iframe that needs additional DOM verification work in headless Chromium. The test gracefully stops at asserting the playground panel opens and the widget root initializes. Can be added in a follow-up.

### Automated (unit tests — `utils.test.ts`)

| Scenario | What's verified |
|----------|----------------|
| `--skip-provider` flag | Scaffolds with `mock/default` model config, skips provider prompt, skips API key prompt |
| Interactive "Skip" selection | Same mock config when user selects "Skip" from the provider menu |
| Precedence over API key flags | `--skip-provider` + `--openai-key` → mock config wins (key ignored) |

### Automated (E2E — `quickstart.test.ts`)

| Scenario | What's verified |
|----------|----------------|
| CLI scaffolding with `--skip-provider` | Directory structure, `.env` contents (no `OPENAI_API_KEY`), `package.json` shape |
| Full stack startup | API server healthy, database setup, auth init |
| `inkeep push` | Project registered in manage DB, API returns correct tenant + project ID |
| Dashboard login | Email/password form submission → redirect to projects page |
| Project navigation | Projects list → project detail page via card click |
| Agent navigation | Agent card click → agent detail page |
| Playground open | "Try it" button → `#playground-pane` visible |
| Widget initialization | `#inkeep-widget-root` element present in DOM |
| Failure diagnostics | Screenshot + URL + HTML captured on any dashboard lap failure |

### Manually QA-tested (not in automated suite)

- [x] `--skip-provider` CLI flow locally — scaffolds with `mock/default`, no API key prompt
- [x] `--skip-provider` + `--openai-key` together — warning logged, mock config used
- [x] Interactive "Skip" option appears as 5th choice with hint text
- [x] E2E dashboard lap passes in CI (full stack — verified across 14 iterative fix commits)
- [x] CI Playwright caching works (cache hit on subsequent runs)

### CI status

All checks green: CI, E2E, auto-format, Vercel previews, PR review.

## Future considerations

- **Playground message send test** — extend dashboard lap to type a message and assert a response from the mock provider. Blocked on shadow DOM / iframe handling in headless Chromium.
- **`--skip-provider` in README** — document the flag for discoverability (surfaced in review, deferred).

---

### DIFF
--- a/packages/create-agents/src/__tests__/e2e/quickstart.test.ts
+++ b/packages/create-agents/src/__tests__/e2e/quickstart.test.ts
@@ -334,7 +334,7 @@ describe('create-agents quickstart e2e', () => {
         console.log('Playground panel is visible');
 
         console.log('Verifying chat widget initialized');
-        await page.locator('pierce/#inkeep-widget-root').waitFor({ timeout: 15000 });
+        await page.locator('#inkeep-widget-root').waitFor({ timeout: 15000 });
         console.log('Chat widget is initialized');
 
         console.log('Dashboard lap complete');


nick-inkeep:
## Review Response

### 1. Empty catch block for screenshots (ACCEPTED)
Fixed in 8933c4c. Now logs the full error:
```typescript
} catch (screenshotError) {
  console.error('Failed to capture screenshot:', screenshotError);
}
```

### 2. Dashboard server output not captured (ACCEPTED)
Fixed in 8933c4c. Added stdout/stderr handlers in `startDashboardServer()` that log errors and readiness signals in CI, matching the existing pattern for the API `devProcess`.

### 3. Interactive "Skip" label is ambiguous (DECLINED)
The "Skip" label was intentionally chosen during spec design — it aligns with how other CLI tools label skip options (e.g., `pnpm create-next-app` uses "Skip"). The `--skip-provider` flag description already explains the mock provider behavior. Adding "(use mock provider)" to the interactive label would make the select option disproportionately long relative to sibling options ("Anthropic", "OpenAI", etc.).

### CI E2E Fix
Also fixed the root cause of the CI failure — `setup-dev:cloud` exits early when its internal migrations fail (CI already applies them from the monorepo root), which skips auth init and the "default" organization is never created. Added a separate `auth init` step that runs regardless of setup-dev:cloud's exit status.

nick-inkeep:
@claude --full-review

miles-kt-inkeep (approved):

PR #2070, opened 2026-02-17
improve performance time on vercel for traces

shagun-singh-inkeep:


shagun-singh-inkeep:
@claude will this improve performance time for loading traces on vercel

PR #2105, opened 2026-02-18
Fix cookie http

robert-inkeep:
Localhost works ok
When using an public http ip, it doesnt work


robert-inkeep:
@claude review again

PR #2127, opened 2026-02-18
Hide evals from openapi docs

dimaMachina:


agents-api/src/openapi.ts
```diff
@@ -46,7 +46,7 @@ export const TagToDescription = {
 
 export function setupOpenAPIRoutes<E extends Env = Env>(app: OpenAPIHono<E>) {
   // OpenAPI specification endpoint - serves the complete API spec
-  app.get('/openapi.json', (c: Context) => {
+  app.get('/openapi.json', (c) => {
```
amikofalvy, line 49: Is the type properly inferred here, Is that why we don't need to provide type explicitly?

shagun-singh-inkeep:
not all eval endpoints need to be deleted in the docs page just the ones related to dataset runs

agents-api/src/openapi.ts
dimaMachina: yes

PR #2135, opened 2026-02-18
feat: user-scoped scheduled triggers

amikofalvy:
## Summary

Enable scheduled triggers to execute with a specific user's identity (`runAsUserId`), allowing user-scoped credential resolution (e.g., per-user OAuth tokens for GitHub, Slack, Jira) during agent execution. Org admins/owners can also delegate triggers to run on behalf of other users in their organization.

## Motivation

Scheduled triggers are the only invocation path that doesn't support user identity. When an agent has tools with `credentialScope: 'user'`, trigger-invoked executions silently skip those tools because no userId flows into the execution context. Users who need automated, recurring agent workflows with their personal OAuth credentials have no supported path — they must either use project-scoped credentials (not always possible) or manually trigger agents via the playground.

User-scoped credentials are already fully plumbed for interactive sessions (chat, playground). This PR closes the gap for scheduled triggers.

## Approach

The implementation adds an optional `runAsUserId` field to the scheduled trigger schema and threads it through all three execution paths (workflow runner, Run Now, Rerun). When present, the execution context is built with `metadata.initiatedBy = { type: 'user', id: runAsUserId }` instead of the default `{ type: 'api_key', id: triggerId }`. Everything downstream — `generateTaskHandler`, `Agent`, credential resolution — already works with a userId, so no changes are needed beyond the trigger layer.

**Permission model:**
- **Self-scheduling:** Any user with `edit` permission can set `runAsUserId` to themselves
- **Admin delegation:** Only org admins/owners can set `runAsUserId` to a different user
- **Target validation:** Target user must have `use` permission on the project (checked at both write-time and execution-time via `canUseProjectStrict`)

**Security hardening (from review):**
- System identifiers (`'system'`, `'apikey:*'`) rejected as `runAsUserId` — prevents bypassing SpiceDB permission checks
- Caller identity validated before delegation checks
- SpiceDB failures in workflow step handled gracefully with actionable error messages

**Legacy behavior preserved:** Triggers with null `runAsUserId` (all existing triggers) continue to execute without a user identity — no migration or backfill needed.

## Architectural decisions

| Decision | Choice | Rationale |
|---|---|---|
| Field name | `runAsUserId` | Clear intent, distinguishes from `createdBy` (audit trail). 1-way door (schema). |
| Permission model | `edit` + self only; org admin/owner for delegation | Self-scheduling is the common case. Delegation requires elevated trust. |
| Runtime permission check | `canUseProjectStrict` before every execution | Access can be revoked between trigger creation and execution. One SpiceDB call per invocation is acceptable for triggers running at most every few minutes. |
| System identifier rejection | Block `'system'` and `'apikey:'` prefixed values | `canUseProjectStrict` bypasses SpiceDB for these identifiers — a user could set `runAsUserId='system'` to bypass all authorization. |
| Missing credentials | Allow creation, fail gracefully at runtime | User-scoped tools without credentials are skipped (not fatal). User may connect credentials later. |
| Run Now/Rerun permissions | Same as scheduling — admin-delegated triggers require admin to manually run | Consistent execution identity model. |
| Execution context | `{ type: 'user', id }` when userId present, `{ type: 'api_key', id }` when absent | Reuses existing type rather than introducing new `delegated_user` type. |

## Changes

**Schema (`agents-core`)**
- Add nullable `runAsUserId` and `createdBy` varchar(256) columns to `scheduled_triggers` table
- Drizzle migration `0010_first_madripoor.sql`

**API routes (`agents-api/domains/manage/routes/scheduledTriggers.ts`)**
- `validateRunAsUserId` helper: rejects system identifiers, checks delegation permissions (admin/owner only for different user), validates target project access via `canUseProjectStrict`
- `validateRunNowDelegation` helper: caller delegation authority for Run Now/Rerun
- Create route: accepts `runAsUserId`, normalizes empty string to null, validates caller identity and permissions, sets `createdBy`
- Update route: same normalization and validation for `runAsUserId` changes
- Run Now: uses trigger's `runAsUserId` for execution, runtime permission check, delegation permission check
- Rerun: same pattern as Run Now
- All route handlers use `OrgRoles.MEMBER` constant (not string literal)

**Execution (`agents-api/domains/run`)**
- `TriggerService.ts`: `executeAgentAsync` accepts optional `runAsUserId`, builds execution context with user identity, adds OTEL span attributes
- `scheduledTriggerSteps.ts`: `executeScheduledTriggerStep` accepts `runAsUserId`, runs `canUseProjectStrict` before execution with try-catch for SpiceDB failures and actionable error messages
- `scheduledTriggerRunner.ts`: passes `currentTrigger.runAsUserId` through workflow execution

**Observability**
- `user.id` and `trigger.run_as_user_id` OTEL span attributes added conditionally when running as a user

**Tests**
- 25 unit tests covering: self-scheduling, admin delegation, target access rejection, system identifier rejection, empty string normalization, legacy behavior, update, GET responses, Run Now (self/admin/legacy), Rerun (self/admin/legacy), delegation logic validation, execution context metadata

**Documentation**
- Updated `agents-docs/content/talk-to-your-agents/triggers/scheduled.mdx` with User-Scoped Execution section, updated Trigger Configuration table (createdBy marked read-only), and updated Run Now/Rerun sections

**Changeset**
- Patch bump for `@inkeep/agents-core` and `@inkeep/agents-api`

## Test plan

_Manual QA verified via code inspection and automated test suite. No browser available._

- [x] **Quality gates**: Typecheck, lint, tests all green
- [x] **Schema migration**: SQL correctly adds nullable columns with no default
- [x] **OpenAPI contract**: Snapshot includes `runAsUserId` and `createdBy` across all trigger schemas
- [x] **Self-scheduling**: Unit test confirms runAsUserId=caller creates trigger (201)
- [x] **Admin delegation**: Unit test confirms runAsUserId!=caller with admin role succeeds (201)
- [x] **Target user access check**: User without project `use` permission gets 400
- [x] **System identifier rejection**: `'system'` and `'apikey:*'` rejected as runAsUserId (400)
- [x] **Empty string normalization**: Empty string runAsUserId treated as null (legacy behavior)
- [x] **Legacy behavior**: Null runAsUserId backward compatible — execution context uses `api_key` type
- [x] **Execution context**: Builds `{ type: 'user', id }` when present, `{ type: 'api_key', id }` when absent
- [x] **OTEL attributes**: `user.id` and `trigger.run_as_user_id` conditionally added to spans
- [x] **Cross-path consistency**: All 3 execution paths (workflow, Run Now, Rerun) pass `runAsUserId`
- [x] **Runtime permission check**: All 3 paths verify `canUseProjectStrict` before execution
- [x] **SpiceDB failure handling**: Workflow step catches errors and returns actionable messages
- [x] **Webhook trigger isolation**: Webhook path does NOT pass `runAsUserId` (out of scope per spec)
- [x] **Run Now/Rerun delegation**: Self, admin-delegated, and legacy paths all verified

### Known gaps
- UI "Run as User" selector deferred to follow-up PR (spec lists as "Should")
- Non-admin 403 only covered by unit-level logic tests (test env hardcodes tenantRole=OWNER)
- No live API testing (dev server not started)

## Future considerations

- **Multi-user fan-out triggers** — one trigger that runs as N different users
- **Webhook trigger user-scoping** — same pattern could apply to webhook triggers
- **Auto-disable on consecutive failures** — when a user loses access or credentials expire
- **UI "Run as" selector** — dropdown populated from project members with credential status indicator

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agents-api/src/__tests__/manage/integration/projectFull.test.ts
+++ b/agents-api/src/__tests__/manage/integration/projectFull.test.ts
@@ -16,6 +16,9 @@ vi.mock('@inkeep/agents-core', async (importOriginal) => {
     removeProjectFromSpiceDb: vi.fn().mockResolvedValue(undefined),
     syncOrgMemberToSpiceDb: vi.fn().mockResolvedValue(undefined),
     canUseProjectStrict: vi.fn().mockResolvedValue(true),
+    getUserById: vi.fn(() =>
+      vi.fn(() => Promise.resolve({ id: 'mock-user', name: 'Mock', email: 'mock@test.com' }))
+    ),
   };
 });
 


agents-api/src/domains/manage/routes/projectFull.ts
```diff
@@ -410,6 +417,48 @@ app.openapi(
...
+        const existingById = new Map(existingTriggers.map((t) => [t.id, t]));
+
+        for (const [triggerId, triggerData] of Object.entries(agentData.scheduledTriggers)) {
+          const existing = existingById.get(triggerId);
+
+          if (existing) {
+            const changed = isScheduledTriggerChanged(triggerData, existing);
+            if (!changed) continue;
+
+            assertCanMutateTrigger({ trigger: existing, callerId, tenantRole });
```
amikofalvy, line 436: The Ito test shows that a non admin can assign another user (the admin to a trigger). @miles-kt-inkeep can you validate this test to see if it is accurate? Timestamp included in the video and Ito link is on the GH PR.

Image

amikofalvy: It looks to me like this situation is already handled by this function: assertCanMutateTrigger

amikofalvy: Separately, it is curious why the run as column is using human unreadable ids instead of the member names which are more useful. We should address this rendering issue if it is real.

packages/agents-core/drizzle/manage/0011_smooth_whizzer.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "scheduled_triggers" ADD COLUMN "run_as_user_id" varchar(256);--> statement-breakpoint
```
amikofalvy, line 1: No fk relationship because it crosses databases, but this means we need to do application level checks when we delete users to also remove these entries as well so that there are no dangling references to non-existent users.

packages/agents-core/drizzle/manage/0011_smooth_whizzer.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE "scheduled_triggers" ADD COLUMN "created_by" varchar(256);
```
amikofalvy, line 2: Is this a userId or a name?

agents-api/__snapshots__/openapi.json
```diff
@@ -6075,6 +6075,11 @@
           "createdAt": {
             "type": "string"
           },
+          "createdBy": {
+            "maxLength": 256,
+            "nullable": true,
+            "type": "string"
+          },
```
amikofalvy, line 6082: question about if this is a userid or a name. the originating schema should define in the description and it will propagate here properly when snapshot is regenerated.

PR #2159, opened 2026-02-19
Derive scope types and WHERE helpers from single source of truth

amikofalvy:
## Summary

- Introduce `scope-definitions.ts` with `SCOPE_KEYS` as the canonical definition of which columns belong to each scope level (`tenant`, `project`, `agent`, `subAgent`)
- TypeScript scope config types (`ProjectScopeConfig`, `AgentScopeConfig`, `SubAgentScopeConfig`) are now computed from `SCOPE_KEYS` via mapped types instead of manually defined
- `scope-helpers.ts` uses a generic `scopedWhere(level, table, scopes)` that iterates over `SCOPE_KEYS`, eliminating per-level column enumeration. Named wrappers (`projectScopedWhere`, etc.) are preserved for ergonomics
- `types/utility.ts` re-exports the derived types to preserve all existing import paths (zero changes needed in consumers)

### Before (3 independent definitions that can drift)

```
manage-schema.ts:   const projectScoped = { tenantId: ..., projectId: ... }
types/utility.ts:   type ProjectScopeConfig = { tenantId: string; projectId: string }
scope-helpers.ts:   and(eq(table.tenantId, ...), eq(table.projectId, ...))
```

### After (1 definition, everything else derived)

```
scope-definitions.ts:   SCOPE_KEYS.project = ['tenantId', 'projectId'] as const
  → types derived:      type ProjectScopeConfig = { [K in ...]: string }
  → WHERE derived:      scopedWhere('project', table, scopes) iterates SCOPE_KEYS.project
```

Applies on top of PR #2151 (scope-aware query helpers + isolation tests).

## Test plan
- [x] All 13 scoping isolation tests pass
- [x] Full monorepo typecheck passes (14/14 tasks)
- [x] Lint passes
- [x] Format passes

### DIFF
--- a/.changeset/miserable-harlequin-dove.md
+++ b/.changeset/miserable-harlequin-dove.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-core": patch
+---
+
+Derive scope config types and WHERE helpers from single source of truth (scope-definitions.ts)
--- a/packages/agents-core/src/data-access/manage/scope-helpers.ts
+++ b/packages/agents-core/src/data-access/manage/scope-helpers.ts
@@ -1,47 +1,46 @@
-import { and, eq } from 'drizzle-orm';
-import type {
-  AgentScopeConfig,
-  ProjectScopeConfig,
-  SubAgentScopeConfig,
-} from '../../types/utility';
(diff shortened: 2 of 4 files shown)


amikofalvy:
✅ Rebased on latest main

PR #2163, opened 2026-02-19
fix: add typeRoots to prevent global type pollution in 3 packages

nick-inkeep:
## Summary

Adds `"typeRoots": ["./node_modules/@types"]` to `tsconfig.json` in three packages (`ai-sdk-provider`, `agents-work-apps`, `agents-api`) to prevent globally-installed `bun-types` from polluting TypeScript's type resolution and causing typecheck failures on developer machines.

## Motivation

Developers with `bun-types` installed globally (at `~/node_modules/bun-types/`) experience typecheck failures because TypeScript's default `typeRoots` behavior walks up the entire directory tree looking for `@types/*` packages. The global `bun-types` augments `typeof fetch` with a non-standard `preconnect` namespace method, causing `vi.fn()` mocks to be incompatible with `typeof fetch` in test files.

This only affects local development — CI passes because runners don't have global `bun-types`. The root cause is documented in [oven-sh/bun#23500](https://github.com/oven-sh/bun/issues/23500).

## Approach

Restrict each affected package's `typeRoots` to only `./node_modules/@types`, preventing TypeScript from scanning parent directories for type packages. This is the TypeScript-recommended mechanism for controlling auto-included `@types/*` packages.

## Architectural decisions

- **`typeRoots` over `types` array**: `typeRoots` restricts *where* TypeScript looks for types without requiring an explicit enumeration of *which* `@types/*` packages to include. The `types` array would require listing every needed type package and updating it whenever dependencies change.
- **Per-package fix**: Each affected `tsconfig.json` gets its own `typeRoots` rather than a shared base config, keeping the fix local and explicit.

## Changes

- **`packages/ai-sdk-provider/tsconfig.json`** — Added `typeRoots` (fixes 7 typecheck errors in test files)
- **`packages/agents-work-apps/tsconfig.json`** — Added `typeRoots` (fixes 2 typecheck errors in `slack/events.test.ts`)
- **`agents-api/tsconfig.json`** — Added `typeRoots` (fixes 3 typecheck errors in test files)

## How to verify

1. Install `bun-types` globally: `cd ~ && npm install bun-types @types/bun`
2. Run `pnpm typecheck` — should pass (previously failed with `preconnect` errors)
3. Run `pnpm test` — all tests still pass
4. Run `pnpm build` — build output unchanged

## Test plan

- Verified `tsc --noEmit` passes for all three packages with the `typeRoots` change
- Verified all tests pass (`vitest --run`) for `ai-sdk-provider` (34 tests)
- Verified build succeeds and `.d.ts` output is unchanged for `ai-sdk-provider`
- Confirmed no other packages in the monorepo reference or extend these tsconfigs
- Confirmed vitest uses its own Vite-based transform independent of `typeRoots`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agents-api/tsconfig.json
+++ b/agents-api/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "typeRoots": ["./node_modules/@types"],
     "verbatimModuleSyntax": true,
     "target": "ES2022",
     "module": "ESNext",
--- a/packages/agents-work-apps/tsconfig.json
+++ b/packages/agents-work-apps/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "typeRoots": ["./node_modules/@types"],
     "verbatimModuleSyntax": true,
     "target": "ES2022",
     "module": "ESNext",


nick-inkeep:
**CI Status Note:** The `Cypress` workflow (run [22177610394](https://github.com/inkeep/agents/actions/runs/22177610394)) fails with `EADDRINUSE: address already in use 0.0.0.0:3000` — a port conflict on the CI runner, not related to this PR's changes. This was confirmed across 2 retries with the same infrastructure error.

The main `CI` workflow (run [22177610339](https://github.com/inkeep/agents/actions/runs/22177610339)) **passed** — including build, typecheck, lint, and all unit/integration tests.

All other checks are green: Vercel deploys, auto-format, pr-review, E2E tests.

PR #2190, opened 2026-02-19
fix: scope sub_agent_skills unique constraint to tenant/project/agent

nick-inkeep:
## Summary

- Fix multi-tenancy bug where the `sub_agent_skills` unique constraint was only on `(sub_agent_id, skill_id)`, causing cross-tenant collisions when different tenants use sub-agents with the same ID and skill combination
- Updated constraint to `(tenant_id, project_id, agent_id, sub_agent_id, skill_id)` with a Drizzle migration

## Problem

Users in a specific project were unable to save skills, getting:
```
Failed query: insert into "sub_agent_skills" ... 
params: posthog,...,reverse-lookup-agent,researching-documentation,...
```

The root cause: the unique constraint `sub_agent_skills_sub_agent_skill_unique` was scoped only to `(sub_agent_id, skill_id)` — meaning if *any* tenant already had a sub-agent with the same ID using the same skill, no other tenant could create that combination.

## Changes

- **`packages/agents-core/src/db/manage/manage-schema.ts`**: Updated unique constraint to include `tenantId`, `projectId`, `agentId`, `subAgentId`, `skillId`
- **`packages/agents-core/drizzle/manage/0010_oval_angel.sql`**: Migration to drop old constraint and create the new tenant-scoped one

## Test plan

- [x] All 86 test files pass (1493 tests)
- [x] Lint clean
- [x] Migration correctly drops old constraint and creates new one
- [ ] Verify on staging that multiple tenants can save the same sub-agent + skill combination


nick-inkeep:
Addressed review feedback in eaa44d20f:

1. **Cross-tenant regression test added** — `subAgentSkills.cross-tenant.test.ts` with two test cases:
   - Verifies different tenants can use the same `(sub_agent_id, skill_id)` combination (the exact scenario that was failing)
   - Verifies same-tenant duplicates are still blocked by the updated constraint
   
2. **Trailing newline fixed** in the migration SQL file.

Both tests pass locally.

packages/agents-core/src/db/manage/manage-schema.ts
```diff
@@ -290,7 +290,13 @@ export const subAgentSkills = pgTable(
       foreignColumns: [skills.tenantId, skills.projectId, skills.id],
       name: 'sub_agent_skills_skill_fk',
     }).onDelete('cascade'),
-    unique('sub_agent_skills_sub_agent_skill_unique').on(table.subAgentId, table.skillId),
+    unique('sub_agent_skills_sub_agent_skill_unique').on(
```
amikofalvy, line 293: @inkeep Does the data access layer for subagent skills need to have the scopes upated?

PR #2214, opened 2026-02-20
fix: use table recreation for sub_agent_skills UNIQUE constraint migration

nick-inkeep:
## Summary

- Replaces the broken `DROP CONSTRAINT + ADD CONSTRAINT` migration with a backup→drop→recreate→restore approach that works on **both Doltgres and PostgreSQL/PGlite**
- Widens the `sub_agent_skills` UNIQUE constraint from `(sub_agent_id, skill_id)` to `(tenant_id, project_id, agent_id, sub_agent_id, skill_id)` — scoping it per-agent
- **Zero data loss** for existing customers (brownfield tested on both engines)

## Problem

Migration `0010_oval_angel.sql` used standard PostgreSQL `ALTER TABLE DROP CONSTRAINT` which fails on Doltgres because it can't find UNIQUE constraints (they're stored as indexes, not in the constraint store). PR #2208 attempted a table recreation fix but created index name collisions on PGlite because it used the same constraint names on the `_new` table while the original still existed.

## Approach

1. Create `sub_agent_skills_backup` (no constraints) → copy all data
2. Drop original `sub_agent_skills` (frees all constraint/index names)
3. Recreate `sub_agent_skills` with correct wider UNIQUE + PK inline
4. Restore data from backup → drop backup
5. Re-add FK constraints and index

This avoids both the Doltgres DROP CONSTRAINT bug and the PGlite per-schema index name collision.

## Test plan

Validated on **2026-02-20** across both database engines:

### PGlite (CI/test engine)
- [x] All 97 test files pass (1525 tests, 0 failures)
- [x] `subAgentSkills.cross-tenant.test.ts` — cross-tenant isolation works
- [x] `client.test.ts` — fresh database with all migrations applied (greenfield)

### Doltgres Docker v0.55.3 (production engine)
- [x] **Greenfield**: All 11 migrations (0000-0010) apply cleanly — 128 statements, 0 errors
- [x] **Brownfield**: 3 existing rows preserved with exact values after migration (zero data loss)
- [x] PK, UNIQUE, FK, and index all correct post-migration
- [x] Duplicate within same agent correctly rejected
- [x] Same sub_agent+skill across different agents correctly allowed
- [x] No backup table left behind

### Alternatives evaluated and rejected
| Approach | Why rejected |
|----------|-------------|
| `DROP CONSTRAINT + ADD CONSTRAINT` | Doltgres bug — can't DROP UNIQUE constraints |
| PR #2208: `CREATE TABLE _new` with same names | PGlite per-schema index name collision |
| `DROP INDEX` on backing index | Doltgres: "table not found" |
| `ALTER TABLE RENAME CONSTRAINT` | Doltgres: "unsupported command type" |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,7 +153,7 @@ jobs:
 
     services:
       doltgres:
-        image: dolthub/doltgresql:latest
+        image: dolthub/doltgresql:0.55.2
         env:
           DOLTGRES_USER: appuser
           DOLTGRES_PASSWORD: password
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -20,7 +20,7 @@ jobs:
 
     services:
       doltgres:
-        image: dolthub/doltgresql:latest
+        image: dolthub/doltgresql:0.55.2
         env:
           DOLTGRES_USER: appuser
           DOLTGRES_PASSWORD: password


amikofalvy:
Duplicate of #2208

PR #2219, opened 2026-02-20
feat: join from workspace — auto-invite Slack users without Inkeep accounts

omar-inkeep:
## Summary

When **Join from Workspace** is enabled for a Slack workspace, users who run `/inkeep link` and don't have an Inkeep account are directed through an accept-invitation flow that creates their account and seamlessly chains into Slack account linking.

## Flow

1. User runs `/inkeep link` in Slack
2. `tryAutoInvite` checks if the workspace has `shouldAllowJoinFromWorkspace` enabled, fetches the user's email from Slack, and verifies they don't already have an Inkeep account
3. If eligible, creates (or reuses) a pending invitation — with `inviterId` and `authMethod` sourced from the organization's `serviceAccountUserId` and `preferredAuthMethod` columns
4. User gets a **Create Account** message in Slack linking to `/accept-invitation/{id}?returnUrl=/link?token=...`
5. After signup + accepting the invitation, they're redirected to `/link?token=...` which completes Slack linking
6. If auto-invite doesn't apply (user already has an account, feature disabled, org not configured), falls back to the standard JWT link flow

### Backend (`agents-work-apps`, `agents-core`)
- **`tryAutoInvite`** in `commands/index.ts` — uses direct DB calls, deduplicates pending invitations, returns invitation info for the accept-invitation URL
- **`createInvitationInDb`** in `organizations.ts` — self-resolves `serviceAccountUserId` and `preferredAuthMethod` from the org record (throws if not configured)
- **`verify-token` handler** in `users.ts` — handles re-linking to different users and unique constraint race conditions
- **`createCreateInkeepAccountMessage`** in `blocks/index.ts` — Slack Block Kit message with 'Create Account' CTA
- **`getOrganizationMemberByEmail`** in `users.ts` data-access — checks org membership by email
- **Schema** — added `preferredAuthMethod` and `serviceAccountUserId` columns to organization table

### Frontend (`agents-manage-ui`)
- **`accept-invitation` page** — supports `returnUrl` search param for post-signup redirect
- **`JoinFromWorkspaceToggle`** — new component with clearer copy, consistent card styling, label doesn't toggle switch

## Gated By

The entire auto-invite flow is gated by the `shouldAllowJoinFromWorkspace` flag on the workspace record, controlled by the `JoinFromWorkspaceToggle` UI.

### DIFF
--- a/.changeset/thundering-sapphire-swift.md
+++ b/.changeset/thundering-sapphire-swift.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Add join-from-workspace toggle and auto-invite flow for Slack users
--- a/.changeset/well-chocolate-scorpion.md
+++ b/.changeset/well-chocolate-scorpion.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-core": minor
+---
+
+Add organization service account, preferred auth method, and Slack workspace join-from-workspace schema support


agents-manage-ui/src/__screenshots__/Form-should-properly-highlight-nested-error-state-1-diff-chromium.png
amikofalvy, line 1: remove this file (rebase on latest main should fix this flaky test, i think)

packages/agents-core/drizzle/runtime/0015_soft_payback.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "organization" ADD COLUMN "preferred_auth_method" text;--> statement-breakpoint
+ALTER TABLE "organization" ADD COLUMN "service_account_user_id" text;--> statement-breakpoint
```
amikofalvy, line 2: In future PR, disallow removal of the organization's Service Account User

PR #2230, opened 2026-02-21
feat(slack): render rich stream events in Slack threads

anubra266:
## Summary

Extends the Slack streaming event handler to render rich agent output — data components, artifacts, citations, tool errors, and summary breadcrumbs — as structured Slack messages. Adds file-upload fallback for oversized payloads and cleans up tool approval state on denial.

## Motivation

Following the tool approval Slack integration (#2149), agents can now produce structured stream events beyond text: `data-component`, `data-artifact`, `data-summary`, and tool lifecycle signals (`tool-input-available`, `tool-output-denied`, `tool-output-error`). Without handling these events, they were silently dropped — resulting in missing content (sources, structured data) and no feedback when tools fail. This PR surfaces all of them in thread context.

## Approach

New block builder functions in `blocks/index.ts` handle each event type and return Slack Block Kit payloads. The streaming loop in `streaming.ts` dispatches to the appropriate builder and either posts a message or uploads a file when the payload exceeds Slack's block character limit (~2900 chars). State is accumulated per stream:

- A `toolCallIdToName` map links `tool-input-available` call IDs to names so that `tool-output-error` events can be labeled correctly.
- Citations are deduplicated by URL and flushed as a single message after the stream ends, rather than mid-stream, to avoid interleaving with text chunks.
- A `richMessageCount` cap (5) prevents flooding a thread when agents emit many data events.
- Tool errors and summary breadcrumb labels are collected and appended to the `streamer.stop()` footer blocks.

## Architectural decisions

**Citations batched at end vs. inline:** Citations are collected across the stream and posted as a single message after `streamer.stop()`. This keeps the main text thread uninterrupted and mirrors how web UIs typically render sources in a footer.

**File upload fallback for large payloads:** JSON payloads >2900 chars and descriptions >2900 chars are uploaded as files (`uploadV2`) rather than truncated. This preserves data fidelity at the cost of a slightly different UX (file attachment vs. inline block). Requires the new `files:write` OAuth scope in the manifest.

**MAX_RICH_MESSAGES = 5:** A hard cap prevents runaway threads when agents emit many data events. Citations are exempt since they're deduplicated and flushed once. This limit is a conservative starting point and can be tuned.

**`toolCallId` on pendingApprovalMessages:** The existing approval tracking array was extended with `toolCallId` so denial events can find and clean up the correct pending entry without linear name matching (names are not guaranteed unique per turn).

## Changes

**`packages/agents-work-apps/src/slack/services/blocks/index.ts`**
- Add `buildToolOutputErrorBlock(toolName, errorText)` — context block with truncated error text
- Add `buildSummaryBreadcrumbBlock(labels)` — breadcrumb trail of summary labels
- Add `buildDataComponentBlocks(component)` — section fields for flat records, JSON code block for nested, overflow to file
- Add `buildDataArtifactBlocks(artifact)` — sources list for citation-type artifacts, header+description for others, overflow to file
- Add `buildCitationsBlock(citations)` — deduplicated sources list, capped at 10 with "N more" suffix
- Add `isFlatRecord` and `findSourcesArray` internal helpers

**`packages/agents-work-apps/src/slack/services/events/streaming.ts`**
- Add `toolCallId` field to `pendingApprovalMessages` entries
- Accumulate `toolCallIdToName`, `toolErrors`, `citations`, `summaryLabels`, `richMessageCount` state per stream
- Handle `tool-input-available` → populate `toolCallIdToName`
- Handle `tool-output-denied` → splice from `pendingApprovalMessages`
- Handle `tool-output-error` → record error for footer rendering
- Handle `data-component` → post blocks or upload file; increment `richMessageCount`
- Handle `data-artifact` → accumulate citations or post/upload artifact; increment `richMessageCount`
- Handle `data-summary` → collect label for breadcrumb
- Append tool error blocks and summary breadcrumb to `streamer.stop()` footer
- Post citations as a follow-up message after stream stop
- Extend completion log with `toolErrorCount`, `citationCount`, `richMessageCount`

**`packages/agents-work-apps/src/slack/slack-app-manifest.json`**
- Add `files:write` OAuth scope for file upload fallback

## How to verify

1. Trigger an agent that emits a `data-component` event — verify a structured Slack block appears in thread.
2. Trigger an agent with citations (`data-artifact` with `type: "citation"`) — verify a "📚 Sources" section appears after the response text.
3. Trigger a tool that fails (`tool-output-error`) — verify "⚠️ ToolName · failed: …" appears in the footer alongside the agent context block.
4. Deny a tool approval — verify the approval message is removed and no lingering state errors appear in logs.
5. Trigger an agent with a large data payload (>2900 chars JSON) — verify a file attachment appears in the thread rather than a truncated block.

## Test plan

- Manually triggered Slack @mention flows against a dev environment agent with each event type.
- Verified `files:write` scope addition does not break existing message-only flows.
- Confirmed citation deduplication by sending two citations with the same URL — only one appears.
- Confirmed `MAX_RICH_MESSAGES` cap by triggering 6+ data events in a single turn — 5 messages posted, remainder silently dropped (warn log emitted).

### DIFF
--- a/packages/agents-work-apps/src/slack/routes/oauth.ts
+++ b/packages/agents-work-apps/src/slack/routes/oauth.ts
@@ -147,6 +147,7 @@ app.openapi(
       'chat:write',
       'chat:write.public',
       'commands',
+      'files:write',
       'groups:history',
       'groups:read',
       'im:history',


packages/agents-work-apps/src/slack/slack-app-manifest.json
```diff
@@ -52,6 +52,7 @@
         "chat:write",
         "chat:write.public",
         "commands",
+        "files:write",
```
amikofalvy, line 55: with respect to `pnpm setup-slack-dev` 

We need to make sure that this permission is included in all dev slack apps as well as prod slack apps.
This means that all i have to do to get this working with my local slack app is to re-run the setup-slack-dev script.

I suggest that we don't have divergent manifest definitions. Currently I think there is one that is statically written here and another procedurally generated one done by the setup-slack-dev script.

amikofalvy: If you make sure to manually update all of the locations which request this new permission I will take care of merging/condensing them into a single source of truth later.

packages/agents-work-apps/src/slack/services/blocks/index.ts
amikofalvy, line 201: There are many similarities between the two surfaces "Slack Work App" and the "Chat widget", Each of them needs to handle the agent event stream. I think it makes sense to explicitly define the set of events through a formal contract. Not blocking the merge of this PR, but I think there is something to improve in the organization of our codebase here. I will proposed a spec in a separate PR for us to review.

packages/agents-work-apps/src/slack/services/events/streaming.ts
```diff
@@ -30,6 +35,30 @@ const CHATSTREAM_OP_TIMEOUT_MS = 20_000;
 /** Shorter timeout for best-effort cleanup in error paths to bound total error handling time. */
 const CLEANUP_TIMEOUT_MS = 3_000;
 
+/**
+ * Retry a file upload on transient failures (5xx, AbortError) with exponential backoff.
+ */
+async function retryUpload<T>(fn: () => Promise<T>, maxAttempts = 3): Promise<T> {
```
amikofalvy: This looks like a general purpose HTTP request retry function. This function is likely already defined and has unit tests. We will be able to re-use it here.

packages/agents-work-apps/src/slack/services/events/streaming.ts
```diff
@@ -243,7 +248,17 @@ export async function streamAgentResponse(params: {
...
+      messageTs: string;
+      toolName: string;
+      toolCallId: string;
+    }> = [];
+    const toolCallIdToName = new Map<string, string>();
+    const toolErrors: Array<{ toolName: string; errorText: string }> = [];
+    const citations: Array<{ title?: string; url?: string }> = [];
+    const summaryLabels: string[] = [];
+    let richMessageCount = 0;
+    const MAX_RICH_MESSAGES = 5;
```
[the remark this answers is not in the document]
amikofalvy: I think it will be easy to hit this limit and it will be a confusing user experience when it occurs. Since the rich messages are silently dropped users will be left with incomplete responses that will never reach the completion state.

I think we should emit an error block kit component when this state is hit and significantly increase the original number for the max cap. Is there some reason to set it as low as 5? In my experiments with block kit, I was showing as many as 12 top level blocks and it was rendering fine on slack client.

If the number of rich elements was too large maybe the agent definition should be fixed to reduce the number of rich messages.

amikofalvy (approved):

PR #2264, opened 2026-02-23
feat(images): Add blob storage capability

mike-inkeep:
Allow storing files in blob storage. Currently supports:
1. local file system (for dev mode)
2. S3 and S3 compatible blob stores (for self hosted)
3. Vercel blob (cloud)

### DIFF
--- a/agents-api/src/domains/run/services/blob-storage/index.ts
+++ b/agents-api/src/domains/run/services/blob-storage/index.ts
@@ -15,7 +15,7 @@ let instance: BlobStorageProvider | null = null;
 /**
  * Returns the configured blob storage provider with precedence:
  * 1) S3 when BLOB_STORAGE_S3_BUCKET is set
- * 2) Vercel when BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN is set
+ * 2) Vercel when BLOB_READ_WRITE_TOKEN is set
  * 3) Local fallback otherwise
  *
  * Stored keys do not encode backend; changing provider strategy after uploads
@@ -25,7 +25,7 @@ export function getBlobStorageProvider(): BlobStorageProvider {
   if (!instance) {
     if (env.BLOB_STORAGE_S3_BUCKET?.trim()) {
       instance = new S3BlobStorageProvider();
-    } else if (env.BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN?.trim()) {
+    } else if (env.BLOB_READ_WRITE_TOKEN?.trim()) {
       instance = new VercelBlobStorageProvider();
     } else {
       instance = new LocalBlobStorageProvider();
--- a/agents-api/src/domains/run/services/blob-storage/vercel-blob-provider.ts
+++ b/agents-api/src/domains/run/services/blob-storage/vercel-blob-provider.ts
@@ -15,7 +15,7 @@ export class VercelBlobStorageProvider implements BlobStorageProvider {
   private token: string | undefined;
 
   constructor() {
-    this.token = env.BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN?.trim() || undefined;
+    this.token = env.BLOB_READ_WRITE_TOKEN?.trim() || undefined;
     logger.info({}, 'Initializing Vercel Blob storage provider');
   }
 
--- a/agents-api/src/env.ts
+++ b/agents-api/src/env.ts
@@ -196,7 +196,7 @@ const envSchema = z
       .describe(
         'Directory path for local blob storage fallback. Resolved relative to process cwd. Default .blob-storage.'
       ),
-    BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: z
+    BLOB_READ_WRITE_TOKEN: z
       .string()
       .optional()
       .describe(
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -61,12 +61,12 @@ INKEEP_AGENTS_TEMP_JWT_PUBLIC_KEY=<REPLACE_WITH_INKEEP_AGENTS_TEMP_JWT_PUBLIC_KE
 # Provider selection order:
 # 1) S3 when BLOB_STORAGE_S3_BUCKET is set
 #    May use any S3 compatible storage (AWS S3, Cloudflare R2, Google Cloud, Backblaze B2, DigitalOcean Spaces, etc.)
-# 2) Vercel Blob when BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN is set
+# 2) Vercel Blob when BLOB_READ_WRITE_TOKEN is set
 # 3) Local filesystem fallback when neither of the above are set. Set BLOB_STORAGE_LOCAL_PATH to override default path for this directory.
 # ===== Local filesystem configuration =====
 BLOB_STORAGE_LOCAL_PATH=.blob-storage
 # ===== Vercel Blob configuration =====
-# BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN=
+# BLOB_READ_WRITE_TOKEN=
 # ===== S3 configuration =====
 # Required for AWS S3:
 # BLOB_STORAGE_S3_BUCKET=
--- a/.env.example
+++ b/.env.example
@@ -149,12 +149,12 @@ INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET=test-bypass-secret-for-ci
 # Provider selection order:
 # 1) S3 when BLOB_STORAGE_S3_BUCKET is set
 #    May use any S3 compatible storage (AWS S3, Cloudflare R2, Google Cloud, Backblaze B2, DigitalOcean Spaces, etc.)
-# 2) Vercel Blob when BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN is set
+# 2) Vercel Blob when BLOB_READ_WRITE_TOKEN is set
 # 3) Local filesystem fallback when neither of the above are set. Set BLOB_STORAGE_LOCAL_PATH to override default path for this directory.
 # ===== Local filesystem configuration =====
 BLOB_STORAGE_LOCAL_PATH=.blob-storage
 # ===== Vercel Blob configuration =====
-# BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN=
+# BLOB_READ_WRITE_TOKEN=
 # ===== S3 configuration =====
 # Required for AWS S3:
 # BLOB_STORAGE_S3_BUCKET=
--- a/agents-api/src/domains/run/services/__tests__/blob-storage-factory.test.ts
+++ b/agents-api/src/domains/run/services/__tests__/blob-storage-factory.test.ts
@@ -39,7 +39,7 @@ describe('blob storage factory', () => {
         BLOB_STORAGE_S3_ACCESS_KEY_ID: 'key',
         BLOB_STORAGE_S3_SECRET_ACCESS_KEY: 'secret',
         BLOB_STORAGE_S3_FORCE_PATH_STYLE: true,
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: 'token',
+        BLOB_READ_WRITE_TOKEN: 'token',
       },
     }));
 
@@ -55,7 +55,7 @@ describe('blob storage factory', () => {
     vi.doMock('../../../../env', () => ({
       env: {
         BLOB_STORAGE_LOCAL_PATH: dir,
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: 'token',
+        BLOB_READ_WRITE_TOKEN: 'token',
       },
     }));
 
--- a/agents-api/src/domains/run/services/__tests__/vercel-provider.test.ts
+++ b/agents-api/src/domains/run/services/__tests__/vercel-provider.test.ts
@@ -19,7 +19,7 @@ describe('VercelBlobStorageProvider', () => {
   it('trims token before passing it to SDK methods', async () => {
     vi.doMock('../../../../env', () => ({
       env: {
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: '  my-token  ',
+        BLOB_READ_WRITE_TOKEN: '  my-token  ',
       },
     }));
     putMock.mockResolvedValueOnce({});
@@ -42,7 +42,7 @@ describe('VercelBlobStorageProvider', () => {
   it('wraps upload errors with key context', async () => {
     vi.doMock('../../../../env', () => ({
       env: {
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: 'token',
+        BLOB_READ_WRITE_TOKEN: 'token',
       },
     }));
     putMock.mockRejectedValueOnce(new Error('Upload failed'));
@@ -61,7 +61,7 @@ describe('VercelBlobStorageProvider', () => {
   it('downloads data and preserves content type', async () => {
     vi.doMock('../../../../env', () => ({
       env: {
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: 'token',
+        BLOB_READ_WRITE_TOKEN: 'token',
       },
     }));
     getMock.mockResolvedValueOnce({
@@ -80,7 +80,7 @@ describe('VercelBlobStorageProvider', () => {
   it('wraps download errors with key context', async () => {
     vi.doMock('../../../../env', () => ({
       env: {
-        BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: 'token',
+        BLOB_READ_WRITE_TOKEN: 'token',
       },
     }));
     getMock.mockResolvedValueOnce({ statusCode: 404, stream: null });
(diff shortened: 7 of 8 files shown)


agents-api/src/domains/run/services/blob-storage/vercel-blob-provider.ts
```diff
@@ -0,0 +1,54 @@
+import { del, get, put } from '@vercel/blob';
```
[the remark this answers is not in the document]
tim-inkeep: @mike-inkeep i think this is fair?

agents-api/src/domains/run/services/blob-storage/tool-result-images.ts
```diff
@@ -0,0 +1,172 @@
+import type { FilePart, Part } from '@inkeep/agents-core';
+import { getLogger } from '../../../../logger';
+
+const logger = getLogger('tool-result-images');
+
+interface ExtractionResult {
+  strippedResult: any;
+  fileParts: FilePart[];
+}
+
```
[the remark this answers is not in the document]
tim-inkeep: i think the test requirement is also fair

agents-api/src/domains/run/services/blob-storage/local-provider.ts
```diff
@@ -0,0 +1,63 @@
...
+import { env } from '../../../../env';
+import { getLogger } from '../../../../logger';
+import type {
+  BlobStorageDownloadResult,
+  BlobStorageProvider,
+  BlobStorageUploadParams,
+} from './types';
+
+const logger = getLogger('local-blob-storage');
+
```
tim-inkeep, line 12: is our plan to potentially support multiple blob storage providers or a BYO for our customers?

mike-inkeep: yes exactly. This PR provides three blob providers:
1. local filesystem (current file) for dev mode
2. S3 compatible blob storage for self-hosted
3. Vercel's blob store service for cloud

agents-api/src/domains/run/services/blob-storage/vercel-blob-provider.ts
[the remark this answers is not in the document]
mike-inkeep: nope, I think Claude didn't really read that doc it linked. I'm guessing it's referring to 1.0.0 SDK docs. This might all be new with 2.0.0 I'm guessing. Vercel is rolling out private blob repos and so you can no longer get a fetch'able URL for blob objects that are stored in private blob repos.

 https://vercel.com/docs/vercel-blob/using-blob-sdk#get
[Get a blob](https://vercel.com/docs/vercel-blob/using-blob-sdk#get-a-blob)
Retrieve blob content as a stream. For private blobs, this is how you deliver files through your functions. For public blobs, you can use this to process blob content server-side.

[get()](https://vercel.com/docs/vercel-blob/using-blob-sdk#get)
TypeScript
Python
get(urlOrPathname, options);
It accepts the following parameters:

urlOrPathname: (Required) A string specifying the URL or pathname of the blob object to retrieve
options: (Required) A JSON object with the following required and optional parameters:

agents-api/src/domains/run/services/blob-storage/tool-result-images.ts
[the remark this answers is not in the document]
mike-inkeep: moving this file to a future PR, this one doesn't need it

agents-api/src/domains/run/services/blob-storage/vercel-blob-provider.ts
```diff
@@ -0,0 +1,54 @@
...
+      );
+    }
+    const data = new Uint8Array(await new Response(result.stream).arrayBuffer());
+    return {
+      data,
+      contentType: result.blob.contentType || 'application/octet-stream',
+    };
+  }
+
+  async delete(key: string): Promise<void> {
```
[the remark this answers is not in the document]
mike-inkeep: [del()](https://vercel.com/docs/vercel-blob/using-blob-sdk#del)
The del method deletes one or multiple blob objects from the Blob store.

Since blobs are cached, it may take up to one minute for them to be fully removed from the Vercel CDN cache.

TypeScript
Python
del(urlOrPathname, options);
 
del([urlOrPathname], options); // You can pass an array to delete multiple blob objects
It accepts the following parameters:

urlOrPathname: (Required) A string or array of strings specifying the URL(s) or pathname(s) of the blob object(s) to delete.
options: (Optional) A JSON object with the following optional parameter:

agents-api/src/domains/run/services/blob-storage/s3-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor() {
+    const bucket = env.BLOB_STORAGE_S3_BUCKET;
+    const accessKeyId = env.BLOB_STORAGE_S3_ACCESS_KEY_ID;
+    const secretAccessKey = env.BLOB_STORAGE_S3_SECRET_ACCESS_KEY;
+    const region = env.BLOB_STORAGE_S3_REGION;
+    if (!bucket || !accessKeyId || !secretAccessKey || !region) {
+      throw new Error(
+        'S3 blob storage requires BLOB_STORAGE_S3_BUCKET, BLOB_STORAGE_S3_ACCESS_KEY_ID, BLOB_STORAGE_S3_SECRET_ACCESS_KEY, and BLOB_STORAGE_S3_REGION'
+      );
+    }
```
amikofalvy, line 36:
```suggestion
    const requiredEnvVars = [
  'BLOB_STORAGE_S3_BUCKET',
  'BLOB_STORAGE_S3_ACCESS_KEY_ID',
  'BLOB_STORAGE_S3_SECRET_ACCESS_KEY',
  'BLOB_STORAGE_S3_REGION',
] as const;

const missing = requiredEnvVars.filter((key) => !env[key]);
if (missing.length > 0) {
  throw new Error(
    `S3 blob storage requires ${missing.join(', ')}`
  );
}
```

We're not even using the scoped variables for bucket, accessKeyId, secretAccessKey, region outside the check for their values being set. In the subsequent lines they are referenced directly from env anyway (i.e. `env.BLOB_STORAGE_S3_ENDPOINT`)

agents-api/src/domains/run/services/blob-storage/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Returns the configured blob storage provider (local, S3, or Vercel). Stored keys do not
+ * encode which backend was used. Do not change BLOB_STORAGE_PROVIDER after uploads
+ * have been made—existing media URLs may 404. Migration between backends is the
+ * deployer's responsibility.
+ */
+export function getBlobStorageProvider(): BlobStorageProvider {
+  if (!instance) {
+    const provider = env.BLOB_STORAGE_PROVIDER ?? 'local';
```
amikofalvy: I think it is unnecessary to have a special env variable for BLOB_STORAGE_PROVIDER when the presence of other env variables can determine what we are trying to use.

To illustrate my point: 

1. If you wanted to use the local provider you would set
```
# common
BLOB_STORAGE_PROVIDER=local

# unique
BLOB_STORAGE_LOCAL_PATH=/tmp
```

2. if you wanted to use S3 you would set:
```
# common
BLOB_STORAGE_PROVIDER=s3

# unique
BLOB_STORAGE_S3_BUCKET=
BLOB_STORAGE_S3_ACCESS_KEY_ID=
BLOB_STORAGE_S3_SECRET_ACCESS_KEY=
BLOB_STORAGE_S3_REGION=
```

3. and for vercel
```
# common
BLOB_STORAGE_PROVIDER=vercel

# unique
BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN=<token here>
```

We can simplify configuration by relying on the presence of the unique ENV variables to pick the right provider and we can remove `BLOB_STORAGE_PROVIDER` entirely.

There might be some future in which the developer my want to use multiple blob storage providers for different purposes, but that is a future problem we can tackle when we get there.

agents-api/src/domains/run/services/blob-storage/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        'Initializing local blob storage provider'
+      );
+      instance = new LocalBlobStorageProvider();
+    } else if (provider === 'vercel') {
+      logger.info({}, 'Initializing Vercel Blob storage provider');
+      instance = new VercelBlobStorageProvider();
+    } else {
+      logger.info({}, 'Initializing S3 blob storage provider');
+      instance = new S3BlobStorageProvider();
+    }
```
amikofalvy: I would put the logger messages inside the constructors of each provider to keep the index file clean of excessive logs

```suggestion
    if (provider === 'local') {
      instance = new LocalBlobStorageProvider();
    } else if (provider === 'vercel') {
      instance = new VercelBlobStorageProvider();
    } else {
      instance = new S3BlobStorageProvider();
    }
```

agents-api/src/env.ts
```diff
@@ -4,182 +4,288 @@ import { loadEnvironmentFiles } from '@inkeep/agents-core';
...
+        data.BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN.trim() === ''
+      ) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: ['BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN'],
+          message:
+            'When BLOB_STORAGE_PROVIDER is "vercel", BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN must be set and non-empty.',
+        });
+      }
+    }
```
amikofalvy, line 263: removing BLOB_STORAGE_PROVIDER will help to clean this up.

We can do a schema validation that at least one of the required env vars needs to be provided if that is necessary to do that? otherwise we can just do per env validation.

agents-api/src/domains/run/services/blob-storage/index.ts
amikofalvy: We will need to pick an order in which we look for the unique variables to resolve situations where more than one ENV is set.

Also, I wouldn't let the default case be vercel, I would make the default case be local provider with default path to simplify local devx

agents-api/src/domains/run/services/blob-storage/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface BlobStorageUploadParams {
+  key: string;
+  data: Buffer | Uint8Array;
```
amikofalvy, line 4: This interface type puts some practical limits on the file size that can be passed through the upload and download interfaces.

This is probably fine for now, but we should figure out what the practical limits are and document them and have a path forward that we can take to do extremely large files of many sizes (and likely support a streaming file interface so we don't have to hold the entire file in memory at any given moment.

Our first version of this does have some additional practical limits on filesize (whatever the LLM providers accept), but I expect that we will want to accept larger files and paginate them (text based documents) in the near future.

agents-api/src/domains/run/services/__tests__/blob-storage-abstraction.test.ts
amikofalvy, line 1: I've recently been focused on the code quality in our test suites.

I want to define some best practices in our codebase for how to write tests.

- When to write unit tests, integration tests.
- How to properly define a mock for easy re-use across multiple files.
- When should we use dynamic imports?
- How should files be broken apart so that the tests for a particular part of code can be easily found without extensive Agentic Search?

In this particular test file, I see that the test for the entire blob-storage module are packaged together into a single test file. I want a reason for this so that we can make a conscious decision on how tests should be done moving forward.

agents-api/src/domains/run/services/blob-storage/index.ts
amikofalvy, line 1: I want to research if this should really be done barrel-file style or if there is a different pattern that we should adopt.

agents-api/src/domains/run/services/blob-storage/local-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+    return resolvedPath;
+  }
+
+  async upload(params: BlobStorageUploadParams): Promise<void> {
+    const filePath = this.keyToPath(params.key);
+    await mkdir(path.dirname(filePath), { recursive: true });
+    const buf =
+      params.data instanceof Buffer ? params.data : Buffer.from(params.data as Uint8Array);
+    await writeFile(filePath, buf);
```
amikofalvy: I recognize that many blob storage providers do override behavior if you attempt to write to a already written key, but since this is local storage for development purposes only should we log a warning if the file already exists?

I wouldn't be too worried about the additional time it takes to check for a local file's presence because this is exclusively meant for local development.

agents-api/src/domains/run/services/blob-storage/local-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+    logger.debug({ key: params.key, path: filePath }, 'Uploaded to local storage');
+  }
```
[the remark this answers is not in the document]
amikofalvy: I recently ran up on a disk full error myself (too many LLM models downloaded locally). I would have benefitted from these error messages. Although, my MB was suffering in other ways as a result of the low disk storage remaining.

agents-api/src/domains/run/services/blob-storage/index.ts
amikofalvy: I suggest:

 1. S3 - if BLOB_STORAGE_S3_BUCKET is set
 2. Vercel - if BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN is set (renamed from BLOB_VERCEL_READ_WRITE_TOKEN)
 3. Local - default fallback (with default path like .blob-storage/)

  Why this order: S3 is the most explicit config (4 required vars), Vercel is single-token, local is zero-config. This makes the common self-hosted path (S3) win when both happen to be set.

agents-api/src/domains/run/services/blob-storage/index.ts
amikofalvy: For blob-storage/, the barrel file is appropriate because:
  - It's an internal service module (not published as a package export) - so tree-shaking concerns don't apply
  - It exports both types and the singleton getter from a coherent module boundary
  - The vercel-react-best-practices skill in this repo warns against barrel files specifically for client-side React code where bundle size matters - not for server-side API code

  Recommendation: Keep the barrel file. It's the right pattern for a server-side service module with a small, cohesive API surface.

agents-api/src/domains/run/services/__tests__/blob-storage-abstraction.test.ts
amikofalvy: Research findings from codebase analysis:

  The codebase uses two test organization patterns:

  Single file per module (dominant pattern): AgentSession.test.ts, ArtifactParser.test.ts, BaseCompressor.test.ts - each tests one class/service.

  Split by concern (used for complex modules): Context validation is split into contextValidation.test.ts, contextValidation.headers.test.ts, contextValidation.integration.test.ts,
  contextValidation.comprehensive.test.ts.

  For blob-storage specifically: The current single file uses vi.resetModules() with dynamic imports to test different provider configurations. This works but has drawbacks:
  - Slow (module re-initialization per test)
  - Hard to find tests for a specific provider
  - Complex mock setup

  Recommendation: Split into:
  blob-storage/__tests__/
    blob-storage-factory.test.ts    # getBlobStorageProvider() singleton, routing logic
    local-provider.test.ts          # Local filesystem provider tests
    s3-provider.test.ts             # S3 provider tests (mocked client)
    vercel-provider.test.ts         # Vercel provider tests (mocked SDK)



I'm deferring the questions/answers to dynamic imports and shared mocks for now.

amikofalvy (changes requested):
```
 Summary of Actionable Items

  ┌─────┬──────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
  │  #  │                   Item                   │                         Verdict                         │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 1   │ Remove BLOB_STORAGE_PROVIDER env var     │ Agree - infer from unique vars, default to local        │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 2   │ S3 env var validation cleanup            │ Agree - use const array + filter pattern                │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 3   │ Move logger to constructors              │ Agree                                                   │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 4   │ Barrel file pattern                      │ Keep as-is - appropriate for server-side service module │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 5   │ Split test file by provider              │ Agree - split into per-provider test files              │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 6   │ Document file size limits                │ Add comment on interface, ~20MB practical limit         │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 7   │ Local overwrite warning                  │ Agree - add logger.warn() on overwrite                  │
  ├─────┼──────────────────────────────────────────┼─────────────────────────────────────────────────────────┤  └─────┴──────────────────────────────────────────┴─────────────────────────────────────────────────────────┘

```

agents-api/src/domains/run/services/blob-storage/types.ts
amikofalvy, line 1: remove file header comments throughout

agents-api/src/env.ts
```diff
@@ -4,182 +4,264 @@ import { loadEnvironmentFiles } from '@inkeep/agents-core';
...
+
+    // Blob Storage (local filesystem fallback, or inferred S3/Vercel)
+    BLOB_STORAGE_LOCAL_PATH: z
+      .string()
+      .optional()
+      .default('.blob-storage')
+      .describe(
+        'Directory path for local blob storage fallback. Resolved relative to process cwd. Default .blob-storage.'
+      ),
+    BLOB_STORAGE_VERCEL_READ_WRITE_TOKEN: z
```
amikofalvy: @pullfrog, it looks like the default env var name for blob storage on Vercel is actually, `BLOB_READ_WRITE_TOKEN`. This needs to be updated here and elsewhere we reference and use this variable

PR #2274, opened 2026-02-24
fix(slack): sanitize citation titles to parse special chars

anubra266:


packages/agents-work-apps/src/slack/services/events/utils.ts
```diff
@@ -90,6 +90,24 @@ export async function findCachedUserMapping(
   return mapping;
 }
 
+/**
+ * Escape special characters in Slack mrkdwn link display text.
+ * In Slack's <url|text> format, `>` terminates the link, `<` opens a new one,
+ * and `&` begins an HTML entity — all must be escaped.
+ */
+export function escapeSlackLinkText(text: string): string {
```
amikofalvy, line 100: We should have unit tests on the two new functions in this file.

amikofalvy (approved):

PR #2278, opened 2026-02-24
updated artifact parsing

tim-inkeep:


amikofalvy (approved):
The changes look good. The new test suite is much needed.

Do an audit to see if there are other character issues that need to be addressed, not just the `>` one that we discovered.

PR #2212, opened 2026-02-20
Unify .env generation between quickstart CLI and contributor flows

nick-inkeep:
## Summary

Unifies `.env` generation between the quickstart CLI (`create-agents`) and the contributor flow (`git clone` + `pnpm setup-dev`). The quickstart CLI previously hardcoded ~30 env vars as a template string and generated 5 secrets inline, diverging from the contributor flow which copies `.env.example` and generates only 2 JWT secrets. This caused:

1. **Missing `INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET`** — broke auto-login and `inkeep push` during `pnpm setup-dev`
2. **Inconsistent defaults** — OTEL/SigNoz pointed at cloud vs local; contributor flow had 3 secrets the CLI generated but `generateJwtKeys()` didn't
3. **Drift** — any new env var added to `.env.example` wasn't picked up by the CLI

Additionally fixes two pre-existing CI e2e test failures:
4. **403 INVALID_ORIGIN on signup/login** — auth API calls used `127.0.0.1` as Origin header, but Better Auth cookies are scoped to `localhost`
5. **Dashboard standalone server not found** — `startDashboardServer()` didn't resolve symlinks when using locally-linked packages via `linkLocalPackages()`

## Architectural decisions

**Copy+inject over hardcoded template string.** `createEnvironmentFiles()` now reads `.env.example` and injects CLI-prompted values (API keys, project ID) by finding and replacing lines. This means any new env var added to `.env.example` is automatically available to CLI users without code changes.

**Secret generation deferred to `generateSecrets()`.** The CLI no longer generates any secrets (RSA keys, signing secrets, auth secrets). All secret generation is consolidated in `generateSecrets()` (renamed from `generateJwtKeys()`), which runs during `pnpm setup-dev`. This ensures both flows use identical secret generation logic.

**Conservative placeholder detection.** `generateSecrets()` only replaces values that are empty, commented out, or match exact placeholder strings (`your-secret-key-change-in-production`, `adminADMIN!@12`). User-customized values are never overwritten.

**localhost for auth, 127.0.0.1 for health checks.** The e2e test has two URL constants: `manageApiUrl` (127.0.0.1 — for health checks and non-auth API calls) and `dashboardApiUrl` (localhost — for auth calls and dashboard env vars). Auth calls must use localhost as Origin because Better Auth's cookie domain is `localhost`.

## Changes

**`create-agents-template/.env.example`**
[...]

PR #2289, opened 2026-02-24
fix(create-agents): sync template deps to CLI version after clone

nick-inkeep:
## Problem

When a user runs `npx @inkeep/create-agents@latest`, the CLI clones `create-agents-template` from GitHub HEAD via `degit`. The template repo has its own `package.json` with `@inkeep/*` dependencies pinned at whatever version was last manually committed — and that version always lags behind the CLI release.

**Example:** CLI publishes v0.52.0, but the template still has `@inkeep/agents-core: "^0.50.3"`. After `pnpm install`, the user ends up with mismatched package versions: the CLI scaffolded code that expects v0.52.0 APIs, but the installed packages are v0.50.x. This causes cryptic runtime errors that are impossible for a new user to diagnose.

The root cause is that template dep versions and CLI versions are updated independently — there's no mechanism to keep them in sync at clone time. This PR fixes that.

## Solution

After cloning the template and `chdir`-ing into the project directory, the CLI now:

1. Reads its own `package.json` version (the version the user is actually running)
2. Rewrites all `@inkeep/*` entries in the template's `package.json` to `^{cliVersion}`
3. Then proceeds to `pnpm install` as before — now with matching versions

This runs before env file generation and install, so the user gets correct deps from the start. If the CLI version can't be read (e.g., running from source), it silently skips the sync — no breaking change.

**Stacked on #2212** — contains only the dep-sync logic, no overlap with env generation changes.

### `packages/create-agents/src/utils.ts` (+35 lines)
- **`getCliVersion()`** — resolves the CLI's own `package.json` via `import.meta.url`, returns the `version` field (or `''` on any failure)
- **`syncTemplateDependencies(templatePath)`** — reads the template's `package.json`, iterates `dependencies` and `devDependencies`, updates any `@inkeep/*` entry to `^{cliVersion}`, writes the file back
- **Call site** — `await syncTemplateDependencies('.')` added in `createAgents()` after `process.chdir(directoryPath)`, before env file generation

### `packages/create-agents/src/__tests__/utils.test.ts` (+161 lines)
- `node:fs` mock (`readFileSync` → returns `{ version: '1.2.3' }`)
- `node:url` mock (`fileURLToPath` → returns `/fake/dist/utils.js`)
- 6 test cases:
  1. Updates `@inkeep/*` deps to `^1.2.3`
  2. Skips gracefully when `package.json` doesn't exist
  3. Handles template with zero `@inkeep/*` deps
  4. Handles template with no `devDependencies` key
  5. Updates `@inkeep/*` in `devDependencies`
  6. Leaves non-`@inkeep` deps (`react`, `next`, etc.) untouched

### `.changeset/meaningful-orange-kangaroo.md`
- Patch: `@inkeep/create-agents`

## What this does NOT touch
- `createEnvironmentFiles()` — handled by #2212
- `createInkeepConfig()` — handled by #2212
- E2E tests — no changes needed
- URL handling / localhost — handled by #2212

## Test plan

- [x] `tsc --noEmit` — passes
- [x] `biome lint --error-on-warnings` — passes
- [x] `biome check` (format) — passes
- [x] 38/38 unit tests pass (26 utils + 12 templates)
- [x] 6 new `syncTemplateDependencies` tests pass
- [ ] CI checks

### DIFF
--- a/packages/create-agents/src/__tests__/utils.test.ts
+++ b/packages/create-agents/src/__tests__/utils.test.ts
@@ -117,6 +117,7 @@ describe('createAgents - Template and Project ID Logic', () => {
     vi.mocked(fs.writeJson).mockResolvedValue(undefined);
     vi.mocked(fs.readJson).mockResolvedValue({});
     vi.mocked(fs.readFile).mockResolvedValue(mockEnvExample as any);
+    vi.mocked(fs.readdir).mockResolvedValue([] as any);
     vi.mocked(fs.mkdir).mockResolvedValue(undefined);
     vi.mocked(fs.remove).mockResolvedValue(undefined);
 
@@ -672,6 +673,7 @@ function setupDefaultMocks() {
   vi.mocked(fs.writeJson).mockResolvedValue(undefined);
   vi.mocked(fs.readJson).mockResolvedValue({});
   vi.mocked(fs.readFile).mockResolvedValue(mockEnvExample as any);
+  vi.mocked(fs.readdir).mockResolvedValue([] as any);
   vi.mocked(getAvailableTemplates).mockResolvedValue(['event-planner', 'chatbot', 'data-analysis']);
   vi.mocked(cloneTemplate).mockResolvedValue(undefined);
   vi.mocked(cloneTemplateLocal).mockResolvedValue(undefined);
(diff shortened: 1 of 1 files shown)


nick-inkeep:
## Review Response

### 🟠 Missing test for `getCliVersion()` failure path — **Accepted**

Valid catch. Added test in `eaf46e621` that overrides `readFileSync` to throw `ENOENT`, then verifies `syncTemplateDependencies` skips without calling `fs.writeJson`. 39/39 tests pass.

### 🟡 Nested workspace packages not synced — **Declined**

Investigated the template structure. The nested packages have:
- `apps/agents-api/package.json` → `@inkeep/agents-api: ^0.50.3`
- `apps/agents-ui/package.json` → `@inkeep/agents-ui: ^0.15.12`

These are in a pnpm workspace (`pnpm-workspace.yaml` → `packages: ["apps/*"]`) with a shared lockfile. Both `^0.50.3` (nested) and `^0.52.0` (root, after sync) resolve to the same version — pnpm resolves deps globally across all workspace packages from one lockfile. There's no actual runtime mismatch, only a cosmetic range difference in the declared `package.json`.

Adding recursive glob/sync logic for all workspace `package.json` files would increase complexity for no functional benefit. If this becomes a real issue (e.g., a workspace package pins an incompatible range), it can be addressed then.

PR #2304, opened 2026-02-24
Tool Result Passing as Arguments (Tool Chaining)

tim-inkeep:
- Agents can pass saved artifacts directly to tools as arguments using a sentinel reference; the system resolves full artifact data (including non-preview fields) before the tool executes — no manual reconstruction needed
  - Agents can chain tool calls by piping the raw output of one tool directly into the next without creating an artifact for intermediate results; primitive return types (strings, numbers, booleans) are fully supported
  - Fixed in-session artifact cache key mismatch that caused getArtifactFull to return the wrong data structure for same-turn artifacts
  - Improved artifact creation prompting to clearly distinguish what artifact:create captures (all fields), what artifact:ref displays (preview only), and what tool argument passing resolves (all fields)


agents-api/src/domains/run/agents/versions/v1/PromptConfig.ts
```diff
@@ -599,10 +599,16 @@ IMPORTANT GUIDELINES:
             : {};
           const fullShape = fullSchema.properties ? buildSchemaShape(fullSchema.properties) : {};
 
-          schemaDescription = `PREVIEW FIELDS — artifact:ref in text or artifact:create (summary only):
+          schemaDescription = `CAPTURED by artifact:create — include ALL of these in your details (both preview and non-preview):
```
amikofalvy: @pullfrog, I'm worried about referecing this syntax as a raw string "artifact:create" throughout multiple places in the repository: prompts, documentation, application business logic. Can we instead extract the agent reference syntaxes at the point of their definition and reference them throughout the codebase?

agents-api/src/domains/run/services/ArtifactParser.ts
```diff
@@ -236,6 +236,19 @@ export class ArtifactParser {
...
+          logger.debug(
+            { toolCallId: args.$tool },
+            'Resolved ephemeral tool result ref in tool arg'
+          );
+          return raw;
+        }
+        logger.warn({ toolCallId: args.$tool }, 'Ephemeral tool result ref could not be resolved');
+        return args;
+      }
+
```
amikofalvy: @claude suggest an improvement to clear up the conditional if statement

agents-docs/content/visual-builder/structured-outputs/artifact-components.mdx
```diff
@@ -126,6 +126,8 @@ Agents can access artifact data in a few ways:
 - **Passing to a tool** — the agent can supply an artifact as an argument to a tool; the tool receives the complete artifact data, including all non-preview fields
 - **On-demand retrieval** — the agent can explicitly fetch the full artifact when it needs to read non-preview fields itself
 
+Not every intermediate tool result needs to become an artifact. When an agent chains tool calls — passing one tool's output directly into the next — no artifact is created and nothing surfaces to the user.
```
amikofalvy, line 130: @pullfrog, looking at this PR as a whole. Does a tool result that gets saved as an artifact get examined for errors or verifying that the shape is the expected structure type before passing to the next tool call?

amikofalvy:
@pullfrog can you handle these 🕐 Pending Recommendations (2)
These items from the https://github.com/inkeep/agents/pull/2304#pullrequestreview-3849373333 remain unaddressed but are Minor severity:

🟡 https://github.com/inkeep/agents/pull/2304#pullrequestreview-3849373333 Missing dedicated regression test for cache key fix (data → full). Tests exist for getToolResultRaw, but no test specifically verifies the cache stores full not data.
🟡 https://github.com/inkeep/agents/pull/2304#pullrequestreview-3849373333 Missing tests for getArtifactFull fallback chain (cache hit → database → raw JSON). The method has good defensive coding but no test coverage for the fallback paths.

agents-docs/content/visual-builder/structured-outputs/artifact-components.mdx
tim-inkeep: @pullfrog can you help us figure out a plan for how we should validate?

tim-inkeep:
@pullfrog, please re-review the WHOLE PR and tell me if all concerns have been alleviated.

PR #2316, opened 2026-02-24
feat(images): End-to-end image persistence and proxying for tool results and user uploads

mike-inkeep:
This is a PR stack which also includes the following PRs:
- feat(images): Persist user images to conversation history https://github.com/inkeep/agents/pull/2372
- feat(images): Resolve blob URLs to media proxy URLs https://github.com/inkeep/agents/pull/2412
- feat(images): Allow tool results to return multi-media content https://github.com/inkeep/agents/pull/2476

**feat(run): Image security for chat API — URL validation, byte sniffing, and format allowlist** 
- Single source of allowed formats: ALLOWED_IMAGE_MIME_TYPES in run/constants/allowed-image-formats.ts (PNG, JPEG, WebP); used by chat schema and blob-storage checks.
- Inline (data URIs): Size limit, MIME allowlist, and byte-sniffing via file-type so claimed type must match actual content.
- External URLs:
  - URL validation (HTTP/HTTPS only, no embedded credentials, allowed ports).
  - DNS resolution check to block private/reserved IPs (localhost, link-local, etc.), including redirect targets.
  - Size limit, redirect limit, and timeout; response body must match allowed image signature.

### DIFF
--- a/packages/agents-core/src/middleware/authz-meta.ts
+++ b/packages/agents-core/src/middleware/authz-meta.ts
@@ -1,8 +1,4 @@
-import type {
-  Env as HonoEnv,
-  Input as HonoInput,
-  MiddlewareHandler,
-} from 'hono';
+import type { Env as HonoEnv, Input as HonoInput, MiddlewareHandler } from 'hono';
 
 export type AuthzMeta = { resource?: string; permission?: string; description: string };
 


agents-api/src/domains/run/constants/allowed-image-formats.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Single source of truth for allowed image formats (chat API and blob-storage security).
+ * Used by run/types/chat.ts (ImageUrlSchema) and run/services/blob-storage image checks.
```
amikofalvy: "Used by [files]" comment should be removed. Any modern IDE has "Find references capability" that allows reverse usage lookup.

agents-api/src/domains/run/services/__tests__/image-security.test.ts
amikofalvy, line 1: Similar to my comment on the previous PR (https://github.com/inkeep/agents/pull/2264#discussion_r2850849994) I'd like to see this test file split up into the individual components for better test separation.

agents-api/src/domains/run/services/blob-storage/external-image-downloader.ts
```diff
@@ -0,0 +1,207 @@
+/**
+ * Downloads an image from a user-provided URL with checks that plain fetch() does not provide.
+ *
+ * Why this exists:
+ * - Blocks requests to private/internal addresses, including redirect targets.
+ * - Enforces strict size limits so very large files cannot exhaust server memory.
+ * - Requires bytes to match an allowed image format instead of trusting headers alone.
+ *
+ * Use this for any untrusted external image URL.
+ */
```
amikofalvy: I'd like to get rid of these file header comments. The filename and the functions contained in the file should be enough to make the purpose of the file clear.

agents-api/src/domains/run/types/chat.ts
amikofalvy, line 21: Should this schema be extracted into core package? The REGEX is re-used, but is the schema or even schema component re-used?

agents-api/src/domains/run/services/blob-storage/image-url-security.ts
```diff
@@ -0,0 +1,111 @@
+/**
+ * URL and DNS guardrails for user-provided URLs.
+ *
+ * This prevents our server from being tricked into calling private/internal services
+ * (for example localhost, metadata endpoints, or internal admin hosts).
+ * Run these checks before any network request to an untrusted URL.
+ */
```
amikofalvy: remove header comment

agents-api/src/domains/run/services/blob-storage/image-security-errors.ts
amikofalvy, line 1: This file exports a bunch of functions that are string interpolation mappers.

My preferred approach would be to extend the Error class with the individual types so that we can provide structured error messages that can be caught and handled appropriately.


So you would have class ImageError extends Error { ... }
InvalidExternalImageUrlError extends ImageError { ...}
and so on.

so that the callers can do
```
try {
  operationThatThrows();
} catch ( error ) {
  if (error instance of InvalidExternalImageUrlError) {
  ...
  }
}
```

amikofalvy: Error types that share a base can be properly formatted by our ApiError handler middleware

agents-api/src/domains/run/services/blob-storage/image-url-security.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (parsed.username || parsed.password) {
+    throw new Error(blockedEmbeddedCredentials);
+  }
+
+  return parsed;
+}
+
+export async function validateUrlResolvesToPublicIp(url: URL): Promise<void> {
+  const hostname = url.hostname;
+  const candidateIps = await resolveCandidateIps(hostname);
```
amikofalvy: Is resolveCandidateIps is expected to throw errors. We should check so that the errors are handled correctly

PR #2319, opened 2026-02-24
feat(dashboard): refactor external agents page to use zod schemas from `agents-core`

dimaMachina:


dimaMachina:

PR #2334, opened 2026-02-25
fix: enable auto-login and inkeep push for create-agents quickstart

nick-inkeep:
## Summary

- Fix auto-login being completely broken for all `create-agents` quickstart users
- Fix `inkeep push` failing with 401 for quickstart users (both during `pnpm setup-dev` and manual runs)

Three independent root causes were identified and fixed:

1. **`proxy.ts` checked `NODE_ENV` instead of `ENVIRONMENT`** — The standalone Next.js build (via `inkeep dev`) always runs with `NODE_ENV=production`, so the proxy's dev-mode guard never passed. Changed to check **both** `ENVIRONMENT` and `NODE_ENV` — this ensures auto-login works for both standalone users (`ENVIRONMENT=development` from `.env`) and monorepo contributors (`NODE_ENV=development` from `next dev`).

2. **`INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET` not injected into generated `.env`** — `createEnvironmentFiles()` reads `.env.example` and injects specific values, but bypass secret was not in the injections map. The generated `.env` kept the template placeholder value (`test-bypass-secret-for-ci`) instead of a unique random secret. Now generates a random 32-byte hex secret via `crypto.randomBytes()` and injects it.

3. **Missing `apiKey` in generated `inkeep.config.ts`** — The CLI config lacked the `apiKey` field, so manual `inkeep push` had no auth credentials. Added `apiKey: process.env.INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET` to match the cookbook template pattern.

## Changes

| File | Change |
|---|---|
| `agents-manage-ui/src/proxy.ts` | Check both `ENVIRONMENT` and `NODE_ENV` for dev-mode guard (handles standalone + monorepo) |
| `packages/create-agents/src/utils.ts` | Generate random bypass secret + add to injections map + add `apiKey` to generated config |

## Environment matrix

| Context | `NODE_ENV` | `ENVIRONMENT` | Auto-login? |
|---|---|---|---|
| **Standalone** (`inkeep dev`) | `production` | `development` (from `.env`) | Yes (via `ENVIRONMENT`) |
| **Monorepo** (`next dev`) | `development` | undefined (not in `agents-manage-ui/.env.example`) | Yes (via `NODE_ENV`) |
| **Production** | `production` | not `development` | No (blocked) |

## Self-fix for existing users

Users who already ran `npx @inkeep/create-agents` can fix their setup:

```bash
# 1. Replace the bypass secret placeholder in .env with a random value
sed -i'' -e "s/INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET=.*/INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET=$(openssl rand -hex 32)/" .env

# 2. Add apiKey to src/inkeep.config.ts (in the agentsApi section):
#    apiKey: process.env.INKEEP_AGENTS_MANAGE_API_BYPASS_SECRET,

# 3. Re-run setup and restart
pnpm setup-dev
pnpm dev
```

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `devSession.test.ts` — all 10 tests pass (API endpoint behavior unchanged)
- [x] `create-agents` tests — all 42 tests pass (bypass secret generation verified)
- [ ] **Manual QA**: Fresh `npx @inkeep/create-agents` → `pnpm setup-dev` → `pnpm dev` → auto-login works
- [ ] **Manual QA**: `inkeep push` succeeds after setup
- [ ] **Regression**: Monorepo `pnpm dev` still auto-logs in

## Spec

[`specs/fix-quickstart-auto-login/SPEC.md`](specs/fix-quickstart-auto-login/SPEC.md)

### DIFF
--- a/packages/create-agents/src/__tests__/e2e/quickstart.test.ts
+++ b/packages/create-agents/src/__tests__/e2e/quickstart.test.ts
@@ -277,8 +277,12 @@ describe('create-agents quickstart e2e', () => {
       }
 
       // --- Dashboard Lap ---
+      // Start the dashboard with ENVIRONMENT=development so the proxy middleware
+      // auto-logs in using the bypass secret — this mirrors the real quickstart
+      // experience where users never see a login form.
       console.log('Starting dashboard lap');
       const dashboardProcess = await startDashboardServer(projectDir, {
+        ENVIRONMENT: 'development',
         INKEEP_AGENTS_API_URL: dashboardApiUrl,
         NEXT_PUBLIC_API_URL: dashboardApiUrl,
         PUBLIC_INKEEP_AGENTS_API_URL: dashboardApiUrl,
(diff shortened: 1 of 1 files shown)


nick-inkeep:
**Re: automated review feedback**

**URL pattern (`127.0.0.1` vs `localhost`):** Intentional. IPv6/IPv4 resolution issues are a real problem on some systems — Node.js may resolve `localhost` to `::1` instead of `127.0.0.1`, causing connection failures. The inline comment explains this. Updating the cookbook template to match is a reasonable follow-up but out of scope for this bug fix.

**Missing `manageUiUrl`:** This field was never in the generated config before this PR. Adding it would be scope expansion. Can be addressed separately.

**Self-fix `>>` append:** Good catch — updated the PR body to use `sed` replacement instead of append, since existing users may already have the placeholder value.

PR #2376, opened 2026-02-25
Remove member count column from channel table, api was changed and we…

sarah-inkeep:
… no longer have this field


robert-inkeep:
@claude add a changeset please

PR #2412, opened 2026-02-26
feat(images): Resolve blob URLs to media proxy URLs

mike-inkeep:
blob_rewrite


agents-api/src/domains/manage/routes/conversations.ts
```diff
@@ -139,4 +159,96 @@ app.openapi(
...
 );
 
+const ConversationMediaParamsSchema = TenantProjectIdParamsSchema.extend({
+  mediaKey: z.string().min(1).openapi({
+    description: 'URL-encoded path segment(s) for the blob (e.g. message key or path)',
+  }),
+});
+
+app.openapi(
+  createProtectedRoute({
```
amikofalvy, line 169: @mike-inkeep I think we may want to have this endpoint return pre-signed urls and referenceIds for the files.

This way our servers don't have to proxy image data through this endpoint.

Clients can fetch images direct from the blob store

PR #2430, opened 2026-02-27
Update dependabot.yml for create-agents-template

robert-inkeep:


robert-inkeep:
@claude change to npm

PR #1669, opened 2026-02-02
chore(dashboard): update vitest deps and add image snapshots for form error state

dimaMachina:
Initial setup of image snapshots, currently Select and Combobox doesn't have red borders when there are validation errors, I'll fix it in following up PR (already fixed in r[euse zod schemas PR](https://github.com/inkeep/agents/pulls))

|  macOS | Linux (CI)  | Pixel mismatch diff <1% |
|---|---|---|
|  image |  image |image |

PR #2439, opened 2026-02-27
chore(dashboard): dockerize Vitest visual regression tests for cross-OS consistency

vnv-varun:
## Summary

- Run Playwright browser inside a Docker container so Vitest screenshot tests produce **identical results** on macOS (local dev) and Linux (CI)
- Replaces the bare `playwright install` in CI with a Docker Playwright server
- Adds `test:visual` / `test:visual:update` npm scripts for local dev
- Fixes Monaco strict locator violation in nested error state visual test

## Problem

Vitest visual/screenshot comparison tests render differently on macOS vs Linux due to OS-level font rendering, anti-aliasing, and subpixel differences. Current workaround (NotoSans font pinning + 2% pixelmatch tolerance) is fragile as more visual tests are added.

**New Docker-rendered baseline (Linux):**

baseline

## Changes

| File | What |
|------|------|
| `docker-compose.visual.yml` | Playwright server container pinned to `linux/amd64` |
| `agents-manage-ui/vitest.config.ts` | `connectOptions` + `exposeNetwork` when `PW_TEST_CONNECT_WS_ENDPOINT` is set |
| `agents-manage-ui/package.json` | `test:visual` and `test:visual:update` scripts |
| `.github/workflows/ci.yml` | Docker Playwright server + health check + env var passthrough |
| `turbo.json` | `passThroughEnv: ["PW_TEST_CONNECT_WS_ENDPOINT"]` for test task |
| `AGENTS.md` | Documented visual test workflow |
| `form.browser.test.tsx` | Fixed Monaco strict locator violation using `data-testid` + Monaco render wait |
| `__screenshots__/*.png` | Baselines regenerated from Linux container |

## How it works

1. Playwright server runs in Docker (Linux container) on port 3100
2. Vitest on the host connects via websocket (`PW_TEST_CONNECT_WS_ENDPOINT`)
3. `exposeNetwork: '<loopback>'` tunnels localhost back to the host so the browser can reach the Vitest dev server
4. All rendering happens in Linux → screenshots are byte-identical across platforms

## Local dev usage

```bash
# Start the Playwright Docker server (one-time, stays running)
docker compose -f docker-compose.visual.yml up -d

# Run visual tests through Docker
pnpm --filter @inkeep/agents-manage-ui test:visual

# Update baselines through Docker
pnpm --filter @inkeep/agents-manage-ui test:visual:update

# Stop when done
docker compose -f docker-compose.visual.yml down
```

## Test plan

- [x] Docker Playwright server starts and accepts connections locally
- [x] All 3 visual tests pass with Docker-rendered baselines
- [x] Vitest `test:visual:update` regenerates baselines from Docker
- [x] Same baselines fail on bare macOS (proves Docker rendering is different)
- [x] CI passes with Docker Playwright server

## Learnings

1. ARM vs x86 Docker renders differently — even same Chromium version, different CPU arch = different subpixel output. Pin platform: linux/amd64 in docker-compose.
2.  Turbo v2 strict mode silently drops env vars — if an env var isn't in env or passThroughEnv in turbo.json, child processes never see it. No error, just silent fallback.
3. Monaco pollutes the DOM with sibling divs — breaks Playwright's auto-generated CSS selectors when using remote browser. Use data-testid for unique locators.

## Related

- Closes [PRD-6191](https://linear.app/inkeep/issue/PRD-6191)
- Closes [PRD-6193](https://linear.app/inkeep/issue/PRD-6193) — Monaco strict locator violation fixed in this PR
- Builds on [#1669](https://github.com/inkeep/agents/pull/1669) by @dimaMachina

### DIFF
--- a/.changeset/brave-foxes-drum.md
+++ b/.changeset/brave-foxes-drum.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Dockerize visual regression tests for cross-OS screenshot consistency


vnv-varun:
### Note on Playwright local install + Docker

We keep `pnpm exec playwright install chromium --with-deps` in CI alongside the Docker Playwright server because `@vitest/browser-playwright` validates that browsers are installed locally before initializing — even when using `connectOptions` to connect to a remote browser.

Without it, CI fails with:
```
Looks like Playwright Test or Playwright was just installed or updated.
Please run the following command to download new browsers:
    pnpm exec playwright install
```

**Research:**
- [Playwright Browsers docs](https://playwright.dev/docs/browsers) — browser install requirements
- [Vitest Playwright connectOptions config](https://vitest.dev/config/browser/playwright) — shows connectOptions usage
- [Vitest Browser Mode docs](https://vitest.dev/guide/browser/) — browser provider setup
- [PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD discussion](https://github.com/microsoft/playwright/issues/10489) — env var to skip downloads doesn't fully work with all setups

The local install is cached in CI and adds minimal overhead. The Docker container is what does the actual rendering — the local install just satisfies the SDK's initialization check.

vnv-varun:
## Learnings

1. **ARM vs x86 Docker renders differently** — even same Chromium version, different CPU arch = different subpixel output. Pin `platform: linux/amd64` in docker-compose.

2. **Turbo v2 strict mode silently drops env vars** — if an env var isn't in `env` or `passThroughEnv` in turbo.json, child processes never see it. No error, just silent fallback.

3. **Monaco pollutes the DOM with sibling divs** — breaks Playwright's auto-generated CSS selectors when using a remote browser server. Use `data-testid` on the container for unique locators.

4. **`@vitest/browser-playwright` requires local browser install even with `connectOptions`** — the provider validates browsers exist locally before connecting to the remote server. Can't skip `playwright install` in CI even though we use Docker. This is arguably a missing feature in vitest ([relevant docs](https://vitest.dev/config/browser/playwright)).

.github/workflows/ci.yml
```diff
@@ -232,7 +253,7 @@ jobs:
         with:
           path: ~/.cache/ms-playwright
           key: playwright-${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: ${{ runner.os }}-playwright-
+          restore-keys: playwright-${{ runner.os }}-
 
       - name: Install Playwright
         run: pnpx playwright install chromium --with-deps
```
dimaMachina:
```suggestion
```

I think we no longer need it?

vnv-varun: I tried this, it seems that @vitest/browser-playwright has a startup check that's overly aggressive: 

When we removed the pnpm exec playwright install chromium --with-deps step from CI, the @vitest/browser-playwright package still needs a local Playwright install to initialize — even though the actual browser runs in Docker. The Playwright npm package validates that browsers are installed locally before connecting.

The fix: we need to keep the Playwright install step (or at least a minimal one) alongside the Docker server. Let me fix this:


Second comment: 

The Playwright npm package has a local validation check — when @vitest/browser-playwright initializes, it checks that Chromium is installed locally on the machine, even if you're going to connect to a remote browser via Docker. If it's not installed, it throws Please run pnpm exec playwright install and refuses to start.

So we need both:

playwright install chromium — satisfies the npm package's local check
Docker Playwright server — where the actual rendering happens
The local install is just for the SDK validation. The browser in Docker is what actually runs the tests. Slightly redundant but that's how Playwright's SDK works.

vnv-varun:
## Restored `onUnhandledError` handler in `vitest.config.ts`

The previous refactor (51f7d70db, 303eb5221) removed the `onUnhandledError` handler from the Vitest config, which caused CI to fail with exit code 1. All 3 tests pass, but without the handler, 3 unhandled Monaco Editor errors bubble up and Vitest treats them as a fatal failure.

**Note:** Ultimately I don't know Monaco well enough here, so if there is a different solution here, I think it should be out of scope for this PR. Ideally a fresh PR -- where these changes can be addressed. Below is my view:

### The errors (all unfixable in Vitest browser mode)

1. **"Cannot use import statement outside a module"** (x2) - Monaco web workers attempt to load ESM modules, but Vitest's browser sandbox doesn't support worker ESM. This is a fundamental incompatibility between Monaco's worker architecture and Vitest browser mode.

2. **"InvalidCharacterError" / "is not a valid name"** - Monaco internally calls `document.createElement()` with an SVG data URI as the tag name. This is Monaco internals, not something we control.

3. **"Closing rpc while"** - A Vitest worker RPC shutdown race condition during test teardown. [Documented Vitest issue](https://github.com/vitest-dev/vitest/issues/9458).

### Why the handler is necessary

- These errors fire during Monaco initialization in the `GenericJsonSchemaEditor` component used by `form.browser.test.tsx`
- The handler only suppresses these 3 specific known error strings - all other unhandled errors propagate normally
- It does NOT use `dangerouslyIgnoreUnhandledErrors` (which would hide everything)
- Without the handler: tests pass but process exits with code 1 (3 errors). With the handler: tests pass, exit code 0.

### History

This handler was originally added by Nick across two PRs after investigation confirmed these are unfixable Monaco/Vitest internals:

- **#2046** - Added `onUnhandledError` with `"Closing rpc while"` suppression for the Vitest worker RPC race condition
- **#2078** - Extended the handler to also suppress `"Cannot use import statement outside a module"` and `InvalidCharacterError` for Monaco web worker errors

### Local verification

Tested both states locally with Docker Playwright:

- **With handler** (`683c53d29` + this fix): 2 files, 3 tests passed, exit code 0, ~7.5s
- **Without handler** (`51f7d70db`): 2 files, 3 tests passed, 3 errors, exit code 1

dimaMachina:
@claude add a changeset

PR #2440, opened 2026-02-27
feat: isolated parallel dev environments

vnv-varun:
## Problem

The current dev setup (`docker-compose.dbs.yml`) uses hardcoded host ports (5432, 5433, etc). Only one dev stack can run at a time. This blocks:
- **Parallel AI agents**: Multiple Claude Code / Ship sessions can't run simultaneously - they all fight over the same ports
- **Preview environments**: Each PR needs its own backing services (Doltgres, Postgres, SpiceDB), impossible with fixed ports
- **Developer productivity**: Can't work on multiple features side-by-side without manually juggling port configs

## Solution

Isolated parallel dev environments with dynamic port allocation at every layer - backing services and application.

**Backing service isolation:**
- `docker-compose.isolated.yml` - same services as `docker-compose.dbs.yml` but with dynamic port allocation (no hardcoded host ports)
- `scripts/isolated-env.sh` - CLI to manage isolated environments (`setup`, `up`, `down`, `status`, `env`)
- `pnpm setup-dev --isolated <name>` - one-command provisioning (Docker + migrations + auth init)
- `.isolated-envs/<name>.json` - port state files for each environment (gitignored)

**App-level isolation:**
- `agents-api/vite.config.ts` - reads `AGENTS_API_PORT` env var (fallback 3002)
- `agents-manage-ui/package.json` - uses `MANAGE_UI_PORT` via shell expansion (fallback 3000)
- `turbo.json` - passes `AGENTS_API_PORT` and `MANAGE_UI_PORT` through to dev processes
- `isolated-env.sh` finds free host ports for both apps via python socket and exports them alongside `INKEEP_AGENTS_API_URL` so manage-ui can reach the correct agents-api
- `setup-dev.js` sets `AGENTS_API_PORT` from state file and uses it for `apiHealthUrl` during setup

**Documentation:**
- `agents-docs/content/community/contributing/isolated-environments.mdx` - full docs on docs.inkeep.com covering quick start, commands, env vars (including app ports), and multi-environment usage
- `AGENTS.md` - updated to reflect app port env vars

## Backwards compatibility

Fully backwards compatible:
- `pnpm setup-dev` (without `--isolated`) uses `docker-compose.dbs.yml` with fixed ports, exactly as before
- `pnpm dev` without env vars starts agents-api on 3002 and manage-ui on 3000, exactly as before
- Only `isolated-env.sh` (or manually setting env vars) activates dynamic app ports

## How it works

1. `COMPOSE_PROJECT_NAME` namespaces everything in Docker (containers, volumes, networks)
2. `ports: - "5432"` (no host binding) lets Docker assign random available ports
3. After startup, backing service ports are discovered via `docker compose port`
4. Free host ports are found for agents-api and manage-ui via `python3 socket.bind(('', 0))`
5. All ports saved to `.isolated-envs/<name>.json`
6. `source <(./scripts/isolated-env.sh env <name>)` exports:
   - Database URLs: `INKEEP_AGENTS_MANAGE_DATABASE_URL`, `INKEEP_AGENTS_RUN_DATABASE_URL`, `SPICEDB_ENDPOINT`
   - App ports: `AGENTS_API_PORT`, `MANAGE_UI_PORT`
   - API URLs: `INKEEP_AGENTS_API_URL`, `PUBLIC_INKEEP_AGENTS_API_URL`, `NEXT_PUBLIC_INKEEP_AGENTS_API_URL`
7. Stale state detection: port probe detects dead containers, cleans up automatically on next `up`

## Usage

```bash
# Full setup (Docker + migrations + auth)
pnpm setup-dev --isolated my-feature

# Point your shell at it
source <(./scripts/isolated-env.sh env my-feature)

# Run app (uses isolated databases + dynamic app ports)
pnpm dev

# Tear down when done
./scripts/isolated-env.sh down my-feature

# See all running environments
./scripts/isolated-env.sh status
```

You can also set app ports manually without isolated environments:
```bash
AGENTS_API_PORT=4002 MANAGE_UI_PORT=4000 pnpm dev
```

## Where this fits

This is a prerequisite for **per-PR preview environments**, where each PR gets its own backing services. The same isolation pattern (dynamic ports, namespace separation) carries forward into the per-PR provisioning workflow.

## Follow-up: Making dynamic ports the default

This PR adds isolated environments as an opt-in alongside the existing fixed-port setup. If we want to make dynamic ports the default (retire `docker-compose.dbs.yml`), there are several hardcoded port references across the codebase that would need updating. That work is tracked in **#2444**.

Tracked in: [PRD-6197](https://linear.app/inkeep/issue/PRD-6197/migrate-default-dev-setup-to-dynamic-port-allocation)

| File | What | Why |
|------|------|-----|
| `packages/agents-core/src/setup/setup.ts:70-73` | `REQUIRED_PORTS` array hardcodes 5432/5433/5434 | Pre-flight port check would fail with dynamic ports. Replace with post-startup health check or skip for dynamic mode. |
| `packages/agents-core/src/auth/authz/config.ts:15` | SpiceDB fallback `localhost:50051` | Env var must always be set - no safe fallback with dynamic ports |
| `agents-api/src/workflow/world.ts:36` | Workflow Postgres fallback `localhost:5432` | Same - env var must always be set |
| `package.json` (3 zed commands) | Hardcoded `--endpoint localhost:50051` | Read from `SPICEDB_ENDPOINT` env var instead |
| `.github/workflows/ci.yml` | Fixed port env vars in CI matrix | CI would need dynamic port discovery or its own isolated setup |
| `.github/workflows/cypress.yml` | Fixed port env vars | Same as ci.yml |
| `.env.example` | Example values reference fixed ports | Update examples and add notes about dynamic mode |

None of these affect this PR (fully backwards compatible), but they're the checklist for a future full migration.

## Test plan

- [x] E2E: `isolated-env.sh setup` - Docker startup, health checks, migrations, auth init
- [x] E2E: `pnpm setup-dev --isolated <name> --skip-push` - full flow completes
- [x] 3 parallel instances + default environment = 16 containers, 20 port assignments, zero collisions
- [x] All instances have identical schema (253 manage tables, 246 runtime tables)
- [x] Clean teardown removes containers + volumes + network; default env unaffected
- [x] Input validation: bad names rejected, `--isolated` without name errors
- [x] Stale state: port probe detects dead containers, cleans up automatically
- [x] Backwards compat: `pnpm setup-dev` (no flag) works exactly as before
- [x] App port defaults: without env vars, agents-api on 3002, manage-ui on 3000
- [x] Custom app ports: `AGENTS_API_PORT=4002 MANAGE_UI_PORT=4000 pnpm dev` overrides correctly
- [x] `cmd_env` exports all 8 env vars (3 DB + 2 app ports + 3 API URLs)
- [x] Status table shows API/UI port columns, backwards compatible with old state files
- [x] `setup-dev.js` uses dynamic port for `apiHealthUrl` (fixed bug: was hardcoded to 3002)

### DIFF
--- a/packages/agents-core/src/setup/setup.ts
+++ b/packages/agents-core/src/setup/setup.ts
@@ -1,6 +1,6 @@
 import { exec, spawn } from 'node:child_process';
 import { generateKeyPairSync, randomBytes } from 'node:crypto';
-import { copyFileSync, createWriteStream, existsSync, writeFileSync } from 'node:fs';
+import { copyFileSync, existsSync, openSync, writeFileSync } from 'node:fs';
 import { readFile, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
@@ -609,9 +609,9 @@ async function startServersIfNeeded(config: SetupConfig): Promise<SpawnedServers
   } else if (config.devApiCommand) {
     logInfo('Starting API server temporarily...');
     const apiLogPath = join(tmpdir(), `setup-dev-api-${Date.now()}.log`);
-    const apiLogStream = createWriteStream(apiLogPath, { flags: 'a' });
+    const apiLogFd = openSync(apiLogPath, 'a');
     const proc = spawn('sh', ['-c', config.devApiCommand], {
-      stdio: ['ignore', apiLogStream, apiLogStream],
+      stdio: ['ignore', apiLogFd, apiLogFd],
       detached: true,
       cwd: process.cwd(),
     });
@@ -629,9 +629,9 @@ async function startServersIfNeeded(config: SetupConfig): Promise<SpawnedServers
     } else if (config.devUiCommand) {
       logInfo('Starting Dashboard temporarily...');
       const uiLogPath = join(tmpdir(), `setup-dev-ui-${Date.now()}.log`);
-      const uiLogStream = createWriteStream(uiLogPath, { flags: 'a' });
+      const uiLogFd = openSync(uiLogPath, 'a');
       const proc = spawn('sh', ['-c', config.devUiCommand], {
-        stdio: ['ignore', uiLogStream, uiLogStream],
+        stdio: ['ignore', uiLogFd, uiLogFd],
         detached: true,
         cwd: process.cwd(),
         env: process.env,


amikofalvy (changes requested):
This PR is missing two things for true isolated parallel dev environments:

### App Level isolation

the changes in this PR isolate the development dependencies (Doltgres, Postgres, SpiceDB), but this PR doesn't cover the application level (agents-api, agents-manage-ui). The new scripts will create an isolated set of docker containers for the downstream dependencies, but the standard `pnpm dev` doesn't create a new set of apps for agents-api and agents-manage-ui.

### Documentation

It looks like way to discover and avoid collisions on ports is script heavy. The modifications to the docker-compose, addition of isolated-env.sh and changes to setup-dev.js are not documented well enough on the highly used docs.inkeep.com. Because these are non-standard development instructions, we need thorough docs to the agents-docs repo. The instructions in Agents.md are well done, but missing the full app level isolation and the written instructions need to be available on https://docs.inkeep.com/community/contributing/overview

vnv-varun (commented):
Hey — both valid points. This PR was intentionally scoped to just the backing services layer (Doltgres, Postgres, SpiceDB) as the foundation. Here's the plan for the rest:

**App-level isolation** — agreed this is needed for true end-to-end parallelism. The isolated backing services here are the prerequisite for the next phase: preview environments ([PRD-6179](https://linear.app/inkeep/issue/PRD-6179)). That workstream provisions full app stacks (agents-api + manage-ui) with Vercel preview deployments pointed at isolated backing services. Specifically:
- [PRD-6168](https://linear.app/inkeep/issue/PRD-6168): Build per-PR Tier 1 provisioning
- [PRD-6169](https://linear.app/inkeep/issue/PRD-6169): Build GitHub Action for preview lifecycle
- [PRD-6170](https://linear.app/inkeep/issue/PRD-6170): E2E test full preview flow

Having isolated backing services with dynamic ports is what makes attaching Vercel preview environments possible — each preview gets its own DB/auth layer without port collisions.

**Documentation** — good point, I've created #2464 to add the isolated env workflow to the contributing docs on docs.inkeep.com.

Happy to consolidate into one PR if you'd prefer, but keeping them separate lets me work on the Vercel preview environment piece and the app-level isolation in parallel. What do you think?

scripts/isolated-env.sh
```diff
@@ -0,0 +1,389 @@
...
+d = json.load(open(sys.argv[1]))
+p = d['ports']
+print(f"export INKEEP_AGENTS_MANAGE_DATABASE_URL='postgresql://appuser:password@localhost:{p['doltgres']}/inkeep_agents'")
+print(f"export INKEEP_AGENTS_RUN_DATABASE_URL='postgresql://appuser:password@localhost:{p['postgres']}/inkeep_agents'")
+print(f"export SPICEDB_ENDPOINT='localhost:{p['spicedb_grpc']}'")
+print(f"export AGENTS_API_PORT='{p['agents_api']}'")
+print(f"export MANAGE_UI_PORT='{p['manage_ui']}'")
+print(f"export INKEEP_AGENTS_API_URL='http://localhost:{p['agents_api']}'")
+print(f"export PUBLIC_INKEEP_AGENTS_API_URL='http://localhost:{p['agents_api']}'")
+print(f"export NEXT_PUBLIC_INKEEP_AGENTS_API_URL='http://localhost:{p['agents_api']}'")
```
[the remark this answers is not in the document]
vnv-varun: Fixed in 7a4681c98. Changed to `.get()` with defaults matching the `cmd_status` pattern:

```python
api = p.get('agents_api', 3002)
ui = p.get('manage_ui', 3000)
```

Also addressed the other review items in the same commit:
- Added `AGENTS_API_PORT`/`MANAGE_UI_PORT` to `.env.example`
- Added warning log for silent port fallback in `setup-dev.js`
- Updated Cypress config + e2e:ci script to respect `MANAGE_UI_PORT`

Tested locally: old state files without `agents_api`/`manage_ui` keys now fall back gracefully instead of crashing.

scripts/setup-dev.js
```diff
@@ -4,29 +4,136 @@
...
+  // Isolated mode: delegate Docker + migrations + auth to isolated-env.sh,
+  // then run the remaining setup steps (secrets, project push) via runSetup.
+  const scriptPath = new URL('./isolated-env.sh', import.meta.url).pathname;
+
+  if (!existsSync(scriptPath)) {
+    console.error(`Error: ${scriptPath} not found`);
+    process.exit(1);
+  }
+
+  console.log(`\n\x1b[1m=== Isolated Environment Setup: ${isolatedName} ===\x1b[0m\n`);
```
dimaMachina: use `import { styleText } from 'node:util';` instead of:
```suggestion
  console.log(
    styleText(
      'bold',
      `\n=== Isolated Environment Setup: ${isolatedName} ===\n`
    )
  );
```

for all inlining styles in console.logs

scripts/setup-dev.js
```diff
@@ -4,29 +4,136 @@
...
+  // remaining steps (secrets, project push) use the right databases.
+  const stateFile = `.isolated-envs/${isolatedName}.json`;
+  if (!existsSync(stateFile)) {
+    console.error(`Error: state file ${stateFile} not found — isolated-env.sh may have failed`);
+    process.exit(1);
+  }
+
+  let state;
+  try {
+    state = JSON.parse(readFileSync(stateFile, 'utf-8'));
```
dimaMachina, line 74: we can import json directly
```suggestion
    state = await import(stateFile, {
      with: {
        type: 'json',
      },
    })
```

vnv-varun:
Small fix pushed in the latest commit - the dynamic `import()` approach for loading the state file has a path resolution quirk in ESM: relative paths resolve against the calling file's location (`scripts/`) rather than `process.cwd()`, so `.isolated-envs/<name>.json` was resolving to `scripts/.isolated-envs/<name>.json` and would throw at runtime.

Reverted to `JSON.parse(readFileSync(...))` which uses the CWD-relative path correctly. The `import()` with JSON assertions is a nice pattern in general, just not the right fit here where the file path is relative to the working directory rather than the module.

vnv-varun:
@claude please go ahead and make your suggestion above

PR #2472, opened 2026-03-02
fix loading `.svg` files as React components in tests 

dimaMachina:


dimaMachina:
@claude add a changeset

PR #2486, opened 2026-03-03
feat: user profiles with timezone storage

miles-kt-inkeep:
## Summary

- Adds a new `user_profile` table (one-to-one with Better Auth's `user` table) with `timezone` (IANA string, nullable) and `attributes` (jsonb, extensible catch-all)
- GET/PUT API endpoints for users to read and update their own profile
- Browser timezone is automatically captured at sign-up and saved to the profile
- Profile row is created via Better Auth `user.create.after` hook on registration
- Profile settings page at `/[tenantId]/settings/profile` with searchable timezone picker

## Test plan

- [ ] Unit tests for data access layer (getUserProfile, upsertUserProfile) pass
- [ ] API routes return correct responses (GET returns empty if no profile, PUT validates IANA timezone)
- [ ] Sign-up flow captures and saves browser timezone
- [ ] Profile settings page renders and allows timezone updates

### DIFF
--- a/.changeset/deep-beige-donkey.md
+++ b/.changeset/deep-beige-donkey.md
@@ -1,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Improve trigger page performance by splitting tab content into separate RSC components with Suspense boundaries and using nuqs for client-side tab switching
--- a/agents-manage-ui/src/app/[tenantId]/projects/[projectId]/triggers/page.tsx
+++ b/agents-manage-ui/src/app/[tenantId]/projects/[projectId]/triggers/page.tsx
@@ -1,44 +1,54 @@
 import type { Metadata } from 'next';
-import FullPageError from '@/components/errors/full-page-error';
+import { Suspense } from 'react';
 import { PageHeader } from '@/components/layout/page-header';
-import { TriggersTabs } from '@/components/project-triggers';
+import { ScheduledTabContent } from '@/components/project-triggers/scheduled-tab-content';
+import { TriggersTabs } from '@/components/project-triggers/triggers-tabs';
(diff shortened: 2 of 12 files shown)


miles-kt-inkeep:
@claude add a changeset

packages/agents-core/src/auth/auth.ts
```diff
@@ -277,6 +278,17 @@ export function createAuth(config: BetterAuthConfig) {
           },
         },
       },
+      user: {
+        create: {
+          after: async (user) => {
```
omar-inkeep: can we move this inside `afterAcceptInvitation`

https://github.com/inkeep/agents/blob/main/packages/agents-core/src/auth/auth.ts#L407

@miles-kt-inkeep

agents-api/src/domains/manage/index.ts
```diff
@@ -13,14 +13,16 @@ import playgroundTokenRoutes from './routes/playgroundToken';
...
 import userProjectMembershipsRoutes from './routes/userProjectMemberships';
 import usersRoutes from './routes/users';
 
 export function createManageRoutes() {
   const app = new OpenAPIHono<{ Variables: ManageAppVariables }>();
 
-  // Mount users routes - organizations and providers endpoints
+  // Mount users routes - organizations, providers, and profile endpoints
   app.route('/api/users', usersRoutes);
+  app.route('/api/users', userProfileRoutes);
```
omar-inkeep, line 25: I think we should make these new routes tenant-scoped?

similar to `userProjectMembershipsRoutes`
https://github.com/inkeep/agents/blob/main/agents-api/src/domains/manage/index.ts#L44-L45

robert-inkeep:
@claude can you comment on how this get's backfilled?

PR #2500, opened 2026-03-03
Fix tool approval denial reason not propagating to LLM or parent agent

anubra266:
## Summary

- When a tool call is denied with a reason, the sub-agent's LLM now receives a clean human-readable string instead of the raw internal sentinel object (`{ __inkeepToolDenied: true, ... }`)
- Denial redirect events are now tracked on the `Agent` instance during execution
- When a delegated sub-agent's task is redirected mid-execution (e.g. user denies `get_coordinates(San Francisco)` with reason "I want tokyo instead"), the delegation response now includes a note so the parent agent knows the task evolved

## Problem

Two bugs in the approval denial flow:

**Bug 1** — Sub-agent LLM received a raw JSON sentinel instead of the reason string, so it would see `{"__inkeepToolDenied":true,"toolCallId":"...","reason":"..."}` and sometimes ignore the redirect.

**Bug 2** — Parent agent had no visibility into user-triggered redirects during delegation. In the trace: `activities-planner` asked `get-coordinates-agent` for San Francisco, the user denied and redirected to Tokyo, `get-coordinates-agent` correctly fetched Tokyo — but `activities-planner` received the Tokyo result without context, got confused by the mismatch, and hallucinated San Francisco's coordinates from training data.

## Changes

- [`Agent.ts`](agents-api/src/domains/run/agents/Agent.ts) — Returns `result.reason` string (not the sentinel object) from `wrapToolWithStreaming` so the AI SDK passes a clean message to the LLM; tracks denial events in `taskDenialRedirects[]`; exposes `getTaskDenialRedirects()` getter
- [`generateTaskHandler.ts`](agents-api/src/domains/run/agents/generateTaskHandler.ts) — Prepends a `[NOTE: The user redirected this task...]` prefix to delegation responses when denials occurred, giving the parent agent full context
- Updated tests to match new behavior
- 

https://github.com/user-attachments/assets/d0133f35-dc50-4f43-8bdb-89c1e09cf826


agents-api/src/domains/run/agents/Agent.ts
```diff
@@ -1016,6 +1035,8 @@ export class Agent {
                       'Tool execution denied by user'
                     );
 
+                    this.recordDenial(toolName, toolCallId, approvalResult.reason);
```
amikofalvy, line 1038: I have a concern about our tool approval flow in general. I bring it up here because this is a continuation of what I think is the problematic pattern.

What we are doing is aggregating the denials in memory. I think all tool related states should be persisted beyond application memory so that we can have potentially long running operations.

Let's discuss tomorrow and either we can fix it in this PR or as a separate PR.

@tim-inkeep is refactoring Agent.ts so I think we should implement persistent storage later

amikofalvy (approved):

agents-api/src/domains/run/agents/Agent.ts
nick-inkeep: Worth thinking about how this is related to durable execution. I.e., are the tool related states persisted via durable execution state primatives (using e.g. Vercel Workflows) **and/or** the runtime DB. 

I shared context with Abraham on durable execution, that's the next project.

PR #2502, opened 2026-03-03
Trace improvements

shagun-singh-inkeep:
-  filter conversations with errors

- message contents should render new lines

- agent name discrepancy

### DIFF
--- a/.changeset/moody-deer-own.md
+++ b/.changeset/moody-deer-own.md
@@ -3,4 +3,4 @@
 "@inkeep/agents-api": patch
 ---
 
-trace improvements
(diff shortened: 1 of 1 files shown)


robert-inkeep:
@claude
a.messageContent?.replaceAll('\\n', '\n')

check this
i think this means we are double encoding it somewhere, or not decoding it right
and it doesnt cover text serialization generally
like even \t or \r would not work here

robert-inkeep:
@claude the content though for my spans for ai assistant response look fine though, the text looks right with all the right line breaks

PR #2506, opened 2026-03-04
feat: support batched approval reponses

anubra266:


amikofalvy (approved):

agents-ui-demo/package.json
amikofalvy, line 1: @claude rename directory `agents-ui` to `agents-ui-demo` to match the package name change

amikofalvy: @claude

There is an issue with changesets not working. I think we need to update the config because we updated the package name from agents-ui to agents-ui-demo. investigate and fix


Some errors occurred when validating the changesets config:
The package or glob expression "@inkeep/agents-ui" is specified in the `ignore` option but it is not found in the project. You may have misspelled the package name or provided an invalid glob expression. Note that glob expressions must be defined according to https://www.npmjs.com/package/micromatch.

amikofalvy:
@claude  ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/agents-ui-demo/package.json

pnpm install and update the lockfile to fix the issue

PR #2521, opened 2026-03-04
ci: auto-update OpenAPI snapshot on PRs

vnv-varun:
## Summary
- When the OpenAPI snapshot is out of date on a PR, CI now auto-regenerates it and commits the update back to the PR branch. The snapshot diff remains visible in the PR for code review.
- Claude code review now also triggers on github-actions[bot] commits (e.g. the auto-updated snapshot).

## Changes
**ci.yml (OpenAPI snapshot auto-update):**
- Job-level `contents: write` permission on the `ci` job only (workflow default stays `read`)
- Checkout uses `github.head_ref` on PRs so we can push back to the branch
- Builds agents-api deps via turbo before running the snapshot update
- Runs `openapi:update-snapshot` on PRs (skips fork PRs)
- If snapshot changed: commits, checks branch still exists, pushes with 3-retry + rebase logic
- Fails explicitly if all push retries are exhausted

**claude-code-review.yml:**
- Added `github-actions[bot]` to the bot allow-list so review triggers on automated commits

## Why
Developers sometimes skip or miss the pre-commit hook that auto-updates the OpenAPI snapshot, causing CI to fail. This catches that case automatically. The snapshot diff is still visible in the PR for reviewers to inspect.

Running Claude review on bot commits means automated changes (like the snapshot update) still get reviewed.

Closes PRD-6226

## Tested
Verified end-to-end on this PR: pushed an intentionally stale snapshot (added a dummy field), CI detected the diff, regenerated the snapshot, and committed the fix back as github-actions[bot] (commit bdefee51). The dummy field was correctly removed.

## Test plan
- [x] Push a stale snapshot to a PR branch
- [x] Verify CI auto-updates the snapshot and commits it back
- [x] Verify the snapshot diff is visible in the PR
- [x] Verify CI passes on main (snapshot steps are skipped)
- [ ] Verify Claude review triggers on the bot's snapshot commit (blocked until claude-code-review.yml change lands on main)

### DIFF
--- a/agents-api/__snapshots__/openapi.json
+++ b/agents-api/__snapshots__/openapi.json
@@ -40241,6 +40241,5 @@
       "description": "Operations for managing Slack workspaces",
       "name": "Workspaces"
     }
-  ],
-  "x-test-stale-v2": "testing path-gated auto-update"
+  ]
 }


vnv-varun:
A couple of notes:

**Snapshot gating:** The snapshot update runs on every PR without path filtering. We could gate it to only run when files that affect the OpenAPI spec change (route files, openapi.ts, createApp.ts), but we'd need to update that list whenever new files start affecting the spec. Open to either approach - keeping it ungated is simpler and only adds ~5-15s.

**Claude review on bot commits:** We discussed having all bot commits on PRs trigger Claude review (not just the current allow-list of inkeep-internal-ci and github-actions). Holding off on that broader change for now since it could increase review costs. Currently only the two bots we care about are allow-listed.

amikofalvy (approved):

PR #2574, opened 2026-03-06
fix(github-work-app): Make gitHub MCP access schema project scoped

miles-kt-inkeep:


miles-kt-inkeep:
@claude add a changeset

PR #2587, opened 2026-03-09
feat(slack-work-app): search user tool

miles-kt-inkeep:


miles-kt-inkeep:
@claude add a changeset

PR #2590, opened 2026-03-09
fix agent flow buttons, use normal-case for left and upper-case for right side

dimaMachina:


dimaMachina:
@claude add a changeset

.changeset/fix-flow-button-casing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inkeep/agents-manage-ui": patch
+---
+
+Fix inconsistent text casing on agent flow buttons
```
dimaMachina:
```suggestion
Make agent flow buttons on left panel normal case and on right panel upper case
```

PR #2606, opened 2026-03-10
Fix mid-generation context compression: accurate context slicing across multiple compression cycles, improved distillation quality, and richer compress

tim-inkeep:
- Track generatedMessagesBaseline across compression cycles so each cycle only processes truly new messages (prevents re-feeding old tool results)
- Call markCompressed() after safeCompress succeeds to avoid corrupting baseline on failure
- Wrap handlePrepareStepCompression body in top-level try/catch for graceful degradation on any compressor error
- Make calculateContextSize public; pass live context size to safeCompress instead of stale initial value
- Mark reused artifacts as processed to avoid redundant DB lookups
- Strengthen distillConversation prompt for better incremental summary preservation
- Fix refreshSummaryArtifactNames to only query artifacts missing a name and to include description from DB
- Improve artifact name/description generator: extract last user message prominently, remove 200-char truncation
- Add COMPRESSION_* constants to SPAN_KEYS in otel-attributes.ts
- Emit compression counts (decisions, next_steps) instead of JSON arrays to avoid high-cardinality OTel attributes
- Add base_context_tokens, generated_tokens, total_context_tokens, trigger_at, and overage span attributes for richer compression traces
- Update compression trace panel UI to show generated/total tokens, context limit, summary tokens, and compression ratio
- Add comprehensive unit tests for MidGenerationCompressor and handlePrepareStepCompression


tim-inkeep:
@pullfrog, please re-review and detail exactly what this PR fixes and if it does a good job and addresses everything without introducing weird code compelxities or behaviors

tim-inkeep:
@pullfrog re-review again

packages/agents-core/src/constants/otel-attributes.ts
```diff
@@ -139,6 +139,30 @@ export const SPAN_KEYS = {
   CONTEXT_BREAKDOWN_THINKING_PREPARATION: 'context.breakdown.thinking_preparation_tokens',
   CONTEXT_BREAKDOWN_CONVERSATION_HISTORY: 'context.breakdown.conversation_history_tokens',
   CONTEXT_BREAKDOWN_TOTAL: 'context.breakdown.total_tokens',
+
+  // Compression telemetry
```
tim-inkeep, line 143: @pullfrog are all these used?

tim-inkeep:
> ### 💭 Consider (1) 💭
> 💭 1) `route.ts:1933-1936` **Remaining `triggerAt` fallback uses keys not in query payload**
> 
> **Issue:** The fallback `COMPRESSION_HARD_LIMIT - COMPRESSION_SAFETY_BUFFER` uses SPAN_KEYS that aren't requested in the query payload (lines 1120-1160), so it will always evaluate to `0 - 0 = 0`.
> 
> **Why:** Extremely low impact — if `COMPRESSION_TRIGGER_AT` is missing (old spans), the UI just shows "0" for trigger threshold. Not worth blocking for.
> 
> **Fix (optional):** Either add these keys to the query, or remove the fallback entirely (accept zeros on old spans). Not blocking.
> 
> **Refs:**

@pullfrog, is this fair?

tim-inkeep:
> ### Analysis: `triggerAt` fallback at `route.ts:1933-1936`
> **The review observation is correct.** The fallback is dead code in practice.
> 
> **The facts:**
> 
> The query payload (lines 1122–1160) requests 10 span keys. `COMPRESSION_HARD_LIMIT` and `COMPRESSION_SAFETY_BUFFER` are **not** among them. Since SigNoz only returns attributes that are explicitly requested in the query, the fallback branch:
> 
> ```ts
> const triggerAt =
>   getNumber(span, SPAN_KEYS.COMPRESSION_TRIGGER_AT, 0) ||
>   getNumber(span, SPAN_KEYS.COMPRESSION_HARD_LIMIT, 0) -
>     getNumber(span, SPAN_KEYS.COMPRESSION_SAFETY_BUFFER, 0);
> ```
> 
> …will always resolve to `0 - 0 = 0` when `COMPRESSION_TRIGGER_AT` is missing.
> 
> **Impact assessment is also correct** — this only affects the UI display for old spans that predate the `COMPRESSION_TRIGGER_AT` attribute. The trigger threshold shows "0" instead of a computed value. No runtime or compression logic is affected.
> 
> **Recommendation:** The cleanest fix would be to remove the dead fallback entirely and just use:
> 
> ```ts
> const triggerAt = getNumber(span, SPAN_KEYS.COMPRESSION_TRIGGER_AT, 0);
> ```
> 
> This makes the intent explicit — old spans without `COMPRESSION_TRIGGER_AT` show 0, which is already what happens today. But this is cosmetic and not blocking.
> 
> [Pullfrog](https://pullfrog.com)  ｜ [View workflow run](https://github.com/inkeep/agents/actions/runs/22918579174/job/66510603412) ｜ Using [Claude Code](https://claude.com/claude-code) ｜ Triggered by [Pullfrog](https://pullfrog.com) ｜ [pullfrog.com](https://pullfrog.com) ｜ [𝕏](https://x.com/pullfrogai)

@pullfrog, can you make this change

tim-inkeep:
@claude do you know why the vercel deployment is failing for agents-manage-ui

PR #2607, opened 2026-03-10
fix(agents-api): Upsert user scoped credential references

miles-kt-inkeep:


miles-kt-inkeep:
@claude add a changeset

PR #2609, opened 2026-03-10
Fix GenericKeyValueInput crash from useFormField context violation

omar-inkeep:
## Summary
- `GenericKeyValueInput` used `FormItem`, `FormLabel`, `FormDescription`, and `FormMessage` outside a `<FormField>` wrapper. These components call `useFormField()` which requires both `FormField` and `FormItem` ancestors.
- This was silently broken until e5cd90f changed the context defaults from `{}` to `null` with proper null checks, exposing the misuse.
- Replaced outer context-dependent components with plain HTML equivalents (matching the pattern from #2595) and wrapped each inner `FormControl` in its own `FormItem` inside its `FormField`.

## Affected pages
- Credential create form (`credential-form.tsx`)
- Credential create for Inkeep Cloud (`credential-form-inkeep-cloud.tsx`)
- Credential edit form (`edit-credential-form.tsx`)

## Test plan
- [ ] Navigate to credentials → new → bearer, verify form renders without errors
- [ ] Add/remove key-value header pairs, verify duplicate key detection still works
- [ ] Edit an existing credential with metadata headers, verify form renders and saves


agents-manage-ui/src/components/form/generic-key-value-input.tsx
```diff
@@ -117,39 +113,41 @@ export function GenericKeyValueInput<
...
-                      onKeyDown={(e) => handleKeyDown(e, index, 'key')}
-                      className={cn(
-                        'flex-1',
-                        isDuplicate && 'ring-2 ring-destructive/50 focus-visible:ring-destructive'
-                      )}
-                      aria-invalid={isDuplicate || undefined}
-                      disabled={disabled}
-                    />
-                  </FormControl>
+                  <FormItem className="flex-1">
```
dimaMachina, line 116:
```suggestion
                  <FormItem className="flex-1">
                    <FormLabel isRequired={isRequired}>{label}</FormLabel>
```

this way is better

PR #2666, opened 2026-03-12
do not truncate mcp name and url in mcp cards

dimaMachina:


dimaMachina:

PR #2670, opened 2026-03-12
ci: replace auto-format push with fail-only check on PRs

amikofalvy:
## Summary
- On PRs, auto-format now runs `pnpm format:check` and **fails** if formatting is wrong, instead of pushing a fix commit that re-triggers all CI checks (~9 min saved per format fix)
- On push to main, the existing format + commit + push behavior is preserved as a safety net
- Clear `::error::` message tells developers to run `pnpm format` locally

## Test plan
- [ ] Open a PR with intentionally unformatted code → workflow should fail with actionable error message
- [ ] Open a PR with correctly formatted code → workflow should pass
- [ ] Merge a PR with unformatted code to main → workflow should auto-fix and push a format commit

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -1,85 +1,45 @@
 name: Auto Format
 
 on:
-  pull_request:
-    branches: ["*"]
   push:
     branches:
       - main
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
-
 jobs:
   auto-format:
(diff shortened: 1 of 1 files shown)


amikofalvy:
Addressed the review feedback — removed the PR trigger entirely instead of duplicating the format check. CI already runs `pnpm format:check` at ci.yml:168, so auto-format now only runs on push to main as a safety net. This dramatically simplifies the workflow (removed ~60 lines of PR-specific logic).

amikofalvy:
Closing — auto-format rarely triggers, and when it does, auto-fixing is better UX than failing and making the dev fix it locally.

PR #2671, opened 2026-03-12
ci: prevent OpenAPI snapshot update from re-triggering checks

amikofalvy:
## Summary
- Replaces the auto-commit-and-push behavior for OpenAPI snapshot updates with a verification step that fails CI if the snapshot is out of date
- Removes `contents: write` permission and git author env vars from the CI job since it no longer pushes commits
- Removes fork PR restriction so snapshot verification runs on all PRs

## Why
The previous behavior pushed a snapshot commit back to the PR branch, which re-triggered all required checks (~9 min wasted). Now CI simply fails with a clear error message telling the developer to update the snapshot locally.

## Test plan
- [ ] Open a PR that changes route files without updating the snapshot — CI should fail with the instructional error message
- [ ] Open a PR that changes route files with the snapshot already up to date — CI should pass with "OpenAPI snapshot is up to date"
- [ ] Open a PR with no route changes — snapshot check is skipped entirely

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,15 +22,11 @@ jobs:
     if: ${{ (github.event_name != 'push' || github.actor != 'github-merge-queue[bot]') && github.head_ref != 'changeset-release/main' }}
     runs-on: ubuntu-32gb
     timeout-minutes: 30
-    permissions:
-      contents: write
     env:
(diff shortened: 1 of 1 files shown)


amikofalvy:
Closing — OpenAPI snapshot auto-push rarely fires, and when it does, auto-fixing saves the dev from running the build locally. The occasional re-trigger cost is worth the better UX.

PR #2675, opened 2026-03-12
ci: combine release and production deploy into single workflow

amikofalvy:
## Summary
- Moved deployment jobs (migrate-databases, deploy-agents-api, deploy-agents-manage-ui, promote) from `vercel-production.yml` into `release.yml` as downstream jobs that run after a successful changesets publish
- Deployment jobs only execute when `changesets/action` outputs `published == 'true'`, skipping snapshot releases
- Checkout refs in the deploy jobs use the release tag output from the release job
- `vercel-production.yml` is retained as a manual-only workflow (`workflow_dispatch`) with a configurable `ref` input for re-deploys/rollbacks

## Motivation
Eliminates the ~1-2 min latency from the GitHub `release:published` event propagation between the two workflows.

## Test plan
- [ ] Verify `release.yml` runs deploy jobs after a real changesets publish on main
- [ ] Verify deploy jobs are skipped on snapshot releases (no changesets to publish)
- [ ] Verify `vercel-production.yml` can be manually dispatched with a custom ref
- [ ] Verify all secrets are available in the `release.yml` workflow context

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,10 +19,16 @@ permissions:
   packages: write
   id-token: write
 
+env:
+  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+
 jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    outputs:
+      published: ${{ steps.changesets.outputs.published }}
+      tag: ${{ steps.release-tag.outputs.tag }}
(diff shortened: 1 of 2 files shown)


amikofalvy:
Closing — 1-2 min savings per release isn't worth the added complexity of maintaining deployment logic in two places. The current event-based separation is cleaner.

PR #2677, opened 2026-03-12
fix(compression): add missing fields to required in distillation schemas to pass OpenAI strict-mode validation

mike-inkeep:
**Summary:**
Discovered while testing the image flow on the customer support → Zendesk delegation: when a ticket with multiple image attachments is fetched with download_attachments: true, the raw tool result balloons to ~238K tokens, pushing the context well past  the mid-generation compression trigger (~106K tokens). 

On every compression attempt, OpenAI rejected the distillation call with: Invalid schema for response_format: 'required' must include every key in properties. Missing 'session_id'.

**Root cause:**
The root cause: three fields in both distillation schemas (session_id, _fallback, and the optional artifacts array) used Zod .optional(), which the AI SDK excludes from the required array when generating the JSON schema for OpenAI's strict structured  output mode. 

The fix changes those fields from .optional() to .nullable(), keeping them in required while still allowing null values. A new test directly resolves the Output.object({ schema }) promise and asserts no property is missing from required, which would have caught all three fields upfront.

### DIFF
--- a/agents-api/src/domains/run/agents/generation/__tests__/ai-sdk-callbacks.test.ts
+++ b/agents-api/src/domains/run/agents/generation/__tests__/ai-sdk-callbacks.test.ts
@@ -34,6 +34,8 @@ function makeCompressor(overrides: Partial<Record<string, any>> = {}): any {
       artifactIds: [],
       summary: {
         type: 'conversation_summary_v1',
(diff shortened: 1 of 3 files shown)


mike-inkeep:
@claude please address the above review comments

PR #2684, opened 2026-03-13
fix: Make OpenTelemetry startup idempotent

mike-inkeep:
Vite hot reload frequently tries to re-run the otel SDK startup, which is not idempotent and leads to a 500 error on all subsequent requests. This requires restarting `pnpm dev`.
The error shows up as: `10:04:53 AM [vite] Internal server error: MetricReader can not be bound to a MeterProvider again.`

This PR makes that operation idempotent by using a global var to keep track of whether otel was started up already.


mike-inkeep:
@claude add export back to defaultSDK

move new NodeSDK() into startOpenTelemetrySDK behind the same guard, or cache the instance on globalThis so repeated module evaluations reuse it

mike-inkeep:
@claude re-review please

mike-inkeep:
@claude please address the following. After addressing `create-agents-template`, decide whether we can remove the export from defaultSDK if nothing needs it anymore

Remaining issues
1. create-agents-template still uses the old pattern — create-agents-template/apps/agents-api/src/instrumentation.ts:19-20 still calls defaultSDK.start() directly. While this won't break (since defaultSDK is still exported), it bypasses the idempotency guard — meaning template users will still hit the MetricReader error on HMR. The template should be updated to use startOpenTelemetrySDK():

// create-agents-template/apps/agents-api/src/instrumentation.ts
const { startOpenTelemetrySDK } = await import('@inkeep/agents-api/instrumentation');
startOpenTelemetrySDK();
[Fix this →](https://claude.ai/code?q=In%20create-agents-template%2Fapps%2Fagents-api%2Fsrc%2Finstrumentation.ts%20on%20branch%20fix%2Fhot_reload_metricreader_error%2C%20replace%20lines%2019-20%20%60const%20%7B%20defaultSDK%20%7D%20%3D%20await%20import(...)%3B%20defaultSDK.start()%3B%60%20with%20%60const%20%7B%20startOpenTelemetrySDK%20%7D%20%3D%20await%20import(%27%40inkeep%2Fagents-api%2Finstrumentation%27)%3B%20startOpenTelemetrySDK()%3B%60&repo=inkeep/agents)

2. Module-level singletons are still re-created on HMR — defaultBatchProcessor, otlpExporter, defaultContextManager, defaultSpanProcessors, etc. (lines 25–83) are re-instantiated on every Vite HMR re-evaluation since they're declared at module scope without a globalThis guard. The cached SDK (via getOrCreateSDK()) still references the original instances, so:

flushBatchProcessor() calls forceFlush() on a new defaultBatchProcessor that has no buffered spans — effectively a no-op after HMR
The old otlpExporter, defaultContextManager, etc. are leaked each reload
This is a dev-only issue (HMR doesn't happen in prod), but it means spans produced after a hot reload may silently fail to flush. Consider either guarding these singletons behind globalThis the same way as the SDK, or deriving defaultBatchProcessor from the cached SDK instance.

3. OtelGlobal type could be stricter (nit) — As pullfrog noted, the current type { [key: symbol]: boolean | NodeSDK } allows assigning the wrong type to either key. A tighter type would be:

type OtelGlobal = {
  [OTEL_SDK_KEY]?: NodeSDK;
  [OTEL_STARTED_KEY]?: boolean;
};
This also eliminates the as NodeSDK cast on line 101.

4. Silent error suppression in catch block (minor) — When the MetricReader error catch fires (line 116-118), there's no log output. Adding a logger.debug or logger.warn here would help distinguish clean idempotency from error-recovery idempotency, consistent with createSafeBatchProcessor() at line 37.

PR #2710, opened 2026-03-16
feat: stream resumption for interrupted conversations

anubra266:
## Summary

Implements the [Vercel AI SDK resume pattern](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams), so users who disconnect mid-stream can return and receive the response seamlessly.

Closes PRD-6312


.changeset/grim-apricot-crow.md
amikofalvy: Is this durable execution or resumable stream. Last we spoke these were relatively independent ideas.

anubra266: Seems it's rebased wrongly, durable commits are somehow here, I'll look into this

amikofalvy: We are also missing a minor changeset for agents-core (db migrations)

packages/agents-core/drizzle/runtime/0025_minor_obadiah_stane.sql
amikofalvy, line 1: migrations have drifted since the PR was first opened, once we have approved this PR we need to rebase on main to get the latest. You will need to resolve migration merge conflicts using the drop regen approach

packages/agents-core/src/data-access/runtime/streamChunks.ts
```diff
@@ -0,0 +1,77 @@
+import { and, eq, gt, sql } from 'drizzle-orm';
+import type { AgentsRunDatabaseClient } from '../../db/runtime/runtime-client';
+import { streamChunks } from '../../db/runtime/runtime-schema';
+
+interface StreamScope {
```
amikofalvy: I think you can just call this conversationScope and move the definition to where the other scopes are defined to keep the scope definitions centralized

packages/agents-core/src/data-access/runtime/streamChunks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const insertStreamChunks =
+  (db: AgentsRunDatabaseClient) =>
+  async (params: StreamScope & { chunks: { idx: number; data: string }[] }) => {
+    if (params.chunks.length === 0) return;
+    await db.insert(streamChunks).values(
+      params.chunks.map((chunk) => ({
+        tenantId: params.tenantId,
+        projectId: params.projectId,
+        conversationId: params.conversationId,
```
amikofalvy, line 29: You can reduce the number of lines in this file if you are able to grab the scopes directly from the params
```suggestion
      ...scopes,
```

packages/agents-core/src/data-access/runtime/streamChunks.ts
```diff
@@ -0,0 +1,77 @@
...
+    ];
+    if (params.afterIdx !== undefined) {
+      conditions.push(gt(streamChunks.idx, params.afterIdx));
+    }
+    return db
+      .select({ idx: streamChunks.idx, data: streamChunks.data, isFinal: streamChunks.isFinal })
+      .from(streamChunks)
+      .where(and(...conditions))
+      .orderBy(streamChunks.idx)
+      .limit(params.limit ?? DEFAULT_CHUNK_BATCH_LIMIT);
```
amikofalvy, line 61: Is it possible for the number of chunks to be greater than this default? do we need better pagination

packages/agents-core/src/data-access/runtime/streamChunks.ts
```diff
@@ -0,0 +1,77 @@
...
+        eq(streamChunks.tenantId, params.tenantId),
+        eq(streamChunks.projectId, params.projectId),
+        eq(streamChunks.conversationId, params.conversationId)
+      )
+    );
+};
+
+export const cleanupExpiredStreamChunks =
+  (db: AgentsRunDatabaseClient) =>
+  async (olderThanMinutes = 5) => {
```
amikofalvy: There could be a lot of chunks potentially and this operation might take longer than expected. Should we cap the number of chunks that are deleted at one time? Is this a real concern?

packages/agents-core/src/validation/schemas.ts
```diff
@@ -357,13 +357,16 @@ export const AgentInsertSchema = createInsertSchema(agents, {
...
-    ResourceIdSchema.clone().openapi({
-      description:
-        'ID of the default sub-agent that handles initial user messages. ' +
-        'Required at runtime but nullable on creation to avoid circular FK dependency. ' +
-        'Workflow: 1) POST Agent (without defaultSubAgentId), 2) POST SubAgent, 3) PATCH Agent with defaultSubAgentId.',
-      example: 'my-default-subagent',
-    }),
+    ResourceIdSchema.clone()
+      .nullable()
+      .optional()
```
amikofalvy: @claude agents nullable and optional? I think we don't want this. How is it relevant to this PR?

agents-api/src/index.ts
```diff
@@ -134,6 +134,19 @@ if (workflowWorld === '@workflow/world-postgres' || workflowWorld === 'local') {
...
+import { cleanupExpiredStreamChunks } from '@inkeep/agents-core';
+import runDbClient from './data/db/runDbClient';
+
+const streamChunkCleanupTimer = setInterval(async () => {
+  try {
+    await cleanupExpiredStreamChunks(runDbClient)();
+  } catch (err) {
+    logger.error({ error: err }, 'Failed to cleanup expired stream chunks');
+  }
+}, 60_000);
```
amikofalvy: @claude  in a serverless environment like Vercel how does this run?

agents-api/src/domains/run/routes/conversations.ts
```diff
@@ -396,4 +400,104 @@ app.openapi(
...
+    conversationId,
+  });
+
+  if (!conversation) {
+    return new Response(null, { status: 204 });
+  }
+
+  if (endUserId && conversation.userId !== endUserId) {
+    return new Response(null, { status: 204 });
+  }
```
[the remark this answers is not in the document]
amikofalvy: How was this resolved? It looks like the authorization check fall through still exists.

agents-api/src/index.ts
amikofalvy: @anubra266 I think we need another method to run cleanup here. The Vercel instances don't usually live 1 minute

amikofalvy: The vercel cron job actually makes sense here.

packages/agents-core/src/data-access/runtime/streamChunks.ts
amikofalvy: 1. This value should be named. We don't want to spend time looking for this timeout later when we are debugging
2. Make this value longer than 5 minutes. This value is independent of the Vercel max_duration time (which varies per plan).
3. Do paginated deletes So we don't get a large DB operation.
4. Run Vercel Cron once per day (if on vercel), 
5. For non Vercel deployments continue using setInterval (only on non vercel deployments)

PR #2744, opened 2026-03-18
Blob storage artifact-data keys, binary sanitizer, and observability stripping

mike-inkeep:
## Summary

Adds an artifact binary sanitizer that offloads inline binary data (images, files) embedded in artifact payloads to blob storage, and strips those payloads from observability spans before they reach the tracing backend. Also introduces a dedicated `artifact-data` storage key category, fixes Vercel file part mapping to use `url` instead of `text`, and adds `inlineExternalPdfUrlParts` to support downloading external PDF URLs as inline bytes.

## Key decisions

**Sanitizer runs at persist time, not at generation time.** Artifact data is sanitized when it is written to storage, so any consumer reading back blob URIs gets a stable reference. An alternative (sanitizing at read time) was not chosen because it would require every read path to do async work.

**Content-hash-based `artifact-data` storage keys.** The key structure is `v1/t_{tenantId}/artifact-data/p_{projectId}/a_{artifactId}/sha256-{hash}.{ext}`. Using a content hash means duplicate binary payloads within the same artifact are deduplicated automatically, and keys are deterministic without needing to generate UUIDs.

**`stripBinaryDataForObservability` is synchronous and non-destructive.** Rather than uploading to blob storage before writing spans (which would add latency and an async dependency to the span-writing hot path), binary parts are replaced with a `[binary data ~N bytes, mimeType: X]` placeholder. This keeps span sizes small without blocking.

**`inlineExternalPdfUrlParts` fetches PDFs before the upload pipeline.** Previously, external PDF URIs in `file` parts were explicitly blocked. This helper downloads them early (in `buildPersistedMessageContent`) and converts to inline `bytes` so the rest of the upload pipeline can process them uniformly.

## Manual QA

### DIFF
--- a/agents-api/src/domains/run/services/blob-storage/artifact-binary-sanitizer.ts
+++ b/agents-api/src/domains/run/services/blob-storage/artifact-binary-sanitizer.ts
@@ -28,7 +28,6 @@ function isInlineBinaryPart(value: unknown): value is InlineBinaryPart {
   return (
     (v.type === 'image' || v.type === 'file') &&
     typeof v.data === 'string' &&
-    v.data.length > 1 &&
     !isBlobUri(v.data) &&
     !v.data.startsWith('http')
   );


tim-inkeep:
@mike-inkeep does this only find binary data within artifacts that have been saved? what about the artifacts themselves being moved to blob storage?

mike-inkeep:
closing in favor of https://github.com/inkeep/agents/pull/2745

PR #2745, opened 2026-03-18
feat(artifacts): binary attachment artifacts and blob-backed artifact hydration

mike-inkeep:
## Summary

  - Registers uploaded user files and binary tool outputs as `binary_attachment` artifacts in the ledger, enabling the conversation history and artifact service to track binary blobs by reference instead of inline.
  - Hydrates blob-backed artifacts on demand when the agent calls `get_reference_artifact`, downloading the binary from storage and returning it as a model-consumable `file` content part (image or file data/URL). Falls back to metadata-only if the download fails.
  - Adds MCP `file` content type support in tool result processing for both model input and conversation history (data + URL variants, images routed as `image-data`/`image-url`).
  - Strips inline binary payloads (base64 blobs) from observability traces, logs, and token estimation to prevent bloat and accidental data leakage into telemetry.

  ## Key decisions

  **Attachment artifact IDs are deterministic when possible.** IDs are derived from the SHA-256 content hash embedded in the blob URI (`attachment_{messageId}_{hash}`), falling back to a positional suffix. This means re-uploads of identical files converge to the same artifact ID within a message.

  **Hydration is on-demand, not eager.** Binary content is only downloaded from blob storage when the agent explicitly calls `get_reference_artifact`. Storing and streaming raw binary data eagerly through every pipeline stage would be expensive and unnecessary in most turns.

  **`get_reference_artifact` is wrapped with `wrapToolWithStreaming` for `toModelOutput` only.** Wrapping is required so the AI SDK calls `toModelOutput` on the tool's return value, which converts blob-backed binary content into multimodal `file`/`image` parts the model can consume.  Without this, the SDK falls back to plain JSON serialization and the model never sees the binary.  `get_reference_artifact` is classified as `isInternalToolForUi` so history persistence and session event recording are skipped — matching the pre-PR behavior and avoiding a wasteful download→re-upload round-trip for content already in blob storage.

  **Compressor waits for pending artifacts before querying.** A race existed where the compressor would look up newly-created artifacts before the async persistence path completed. `waitForPendingArtifacts()` (extracted from session cleanup) is now called before the DB read in `buildToolCallArtifactMap`.

  **Artifact retry logic simplified.** The counter-based `MAX_ARTIFACT_RETRIES` + error map approach was replaced with a single fixed-delay retry (`ARTIFACT_SAVE_RETRY_DELAY_MS = 3s`). The previous logic tracked errors across retries but didn't actually schedule retries — it just incremented a counter and logged.

  **Cumulative summary artifact lookup fixed.** The summary map was keyed by `toolCallId` but looked up by `artifactId` when refreshing names/descriptions. The key is now `artifactId` throughout.

  ## Manual QA


agents-api/src/domains/run/agents/tools/default-tools.ts
```diff
@@ -133,7 +237,13 @@ export async function getDefaultTools(
 
   const compressionConfig = getModelAwareCompressionConfig();
   if ((await agentHasArtifactComponents(ctx)) || compressionConfig.enabled) {
-    defaultTools.get_reference_artifact = getArtifactTools(ctx);
+    defaultTools.get_reference_artifact = wrapToolWithStreaming(
```
mike-inkeep, line 253: we are wrapping this so that it goes through to toModelOutput → buildToolResultForModelInput

agents-api/src/domains/run/agents/tools/tool-wrapper.ts
```diff
@@ -105,7 +106,10 @@ export function wrapToolWithStreaming(
       const isInternalTool =
         toolName.includes('save_tool_result') || toolName.startsWith('transfer_to_');
       const isInternalToolForUi =
-        isInternalTool || toolName.startsWith('delegate_to_') || toolName === 'load_skill';
+        isInternalTool ||
```
tim-inkeep, line 139: @mike-inkeep i think previously we had this showing up in the UI for traces, will this block it from appearing for traces? or just for the stream is that what we want? not sure but i can see the argument for hiding it

agents-api/src/domains/run/compression/BaseCompressor.ts
```diff
@@ -93,26 +120,32 @@ export abstract class BaseCompressor {
             total += this.estimateTokens(block.text || '');
           } else if (block.type === 'tool-call') {
```
tim-inkeep, line 126: how are compression summaries going to be comprehensive without binary payloads? will we have a description in their place to demonstrate what was there? is the artifact name and description going to be sufficient? It usually does not generate in-time before a compression result is created and used.

agents-api/src/domains/run/session/AgentSession.ts
```diff
@@ -647,29 +665,10 @@ export class AgentSession {
...
-          {
-            sessionId: this.sessionId,
-            pendingCount: this.pendingArtifacts.size,
-            pendingIds: Array.from(this.pendingArtifacts),
-          },
-          'Cleanup proceeding with pending artifacts still processing'
-        );
-      }
-    }
+    await this.waitForPendingArtifacts();
```
tim-inkeep, line 690: good abstraction

tim-inkeep (commented):
A few things I noticed here:

1. **`get_reference_artifact` in `isInternalToolForUi` hides it from traces**

   `isInternalToolForUi` currently affects both stream visibility and conversation history persistence. Adding `get_reference_artifact` there means it won’t show up in traces at all, not just in the user-facing chat stream.

   If the goal is to hide it from the UI but still keep it available for debugging, this probably needs a more granular flag.

2. **Compression summaries may lose useful context for binary artifacts**

   When compression replaces a tool result with a summary, it relies on the artifact’s `name` and `description`. But those are generated asynchronously in `processArtifact()`, and `waitForPendingArtifacts()` plus the DB refresh can happen before that work finishes.

   In that case, summaries for binary artifacts may fall back to raw metadata like filename / MIME type instead of something more meaningful.

3. **`createAttachmentArtifacts` errors can get swallowed**

   Right now file upload and artifact creation share the same `try/catch` in `buildPersistedMessageContent`. That means if upload succeeds but artifact creation fails later, for example due to a DB error, it gets handled the same way as a `FileSecurityError` and quietly falls back to text-only.

   The result is a file in blob storage with no artifact record, and the message gets saved as if there was never an attachment. Artifact creation failures should probably propagate instead of being grouped with file security failures.

4. **Artifacts can be orphaned if message creation fails**

   All three routes create artifacts inside `buildPersistedMessageContent` before calling `createMessage()`. If the message write fails after the artifacts are already persisted, those artifact rows end up pointing at a message that doesn’t exist.

   Not urgent, but this seems like it could slowly accumulate orphaned records over time.

5. **`executions.ts` is missing file-specific error handling**

   `chat.ts` and `chatDataStream.ts` both have explicit handling for `FileSecurityError` and `PdfUrlIngestionError` around the attachment path. `executions.ts` now goes through the same file attachment flow, but doesn’t have equivalent handlers.

   That means a malicious file or bad PDF URL through this endpoint would likely surface as an unhandled 500 instead of a clean 400.

6. **No size guard when hydrating blob data in `get_reference_artifact`**

   `makeHydratedReferenceArtifactResult` downloads the full blob and base64-encodes it in memory without any size limit. For large stored files, that could create pretty significant memory pressure.

   Seems worth adding a size cap before downloading or hydrating.

agents-api/src/domains/run/agents/tools/default-tools.ts
```diff
@@ -1,19 +1,122 @@
...
+          type: 'text',
+          text: JSON.stringify(metadataContent),
+        },
+      ],
+    };
+  }
+
+  const storage = getBlobStorageProvider();
+  try {
+    const blob = await storage.download(fromBlobUri(artifactData.data.blobUri));
```
amikofalvy, line 61: @mike-inkeep with s3 and presigned urls, can we deliver certain types of files via the presigned url instead of having to download them and re-upload them?

agents-api/src/domains/run/compression/BaseCompressor.ts
mike-inkeep: I've updated it so that this is what's passed in for binary artifacts:
```
  [BINARY ATTACHMENT] read_image [ARTIFACT: attachment_msg1_abc123] "Tool attachment 1"
  Description: Binary file produced by tool (image/png)
  mimeType: image/png, binaryType: image, filename: photo.png
```
That way the distillation LLM now knows what the file was, what format it's in, and can reference it by ID without ever touching the binary payload.

agents-api/src/domains/run/agents/tools/tool-wrapper.ts
mike-inkeep: @tim-inkeep Just updated this code. It'll appear in traces in the UI and in otel, and it'll persist to conversation history. I've set it to not show a UI block in the main chat widget.

mike-inkeep:
@tim-inkeep I've addressed 1 and 2, see my inline comments above. For the rest:

1. createAttachmentArtifacts errors can get swallowed

These are two steps that are both necessary to persist an artifact (write to blob and write to DB). If either fails then we drop the binary part and just return the text part. We do this instead of throwing because any tool call may return binary data at any time, and we want to make a best effort at persisting it to artifact, without blowing up that generation step if it fails.


2. Artifacts can be orphaned if message creation fails

Ditto. This is fine


3. executions.ts is missing file-specific error handling

Same as above. Tool calls returning bad binary data should not blow up the entire generation. We just drop them silently.


4. No size guard when hydrating blob data in get_reference_artifact

We do a size guard when uploading. Anything that made it into blob storage should be considered safe-ish.

PR #2754, opened 2026-03-18
Resolve sentinel refs before tool execution

mike-inkeep:
The AI SDK validates tool args against the tool schema **before** `execute()` is called, so the `$tool`/`$artifact`/`$path` sentinel keys are rejected before `resolveArgs` can replace them with real data.

1. Model outputs `image_info({ image: { $tool: "...", $path: "..." } })`
2. AI SDK validates this against the MCP-derived Zod schema (`z.object({ data: z.string(), mimeType: z.string() })`) — **fails** because `data` and `mimeType` are missing
3. AI SDK throws `InvalidToolInputError`, catches it internally, and marks the tool call as `invalid: true`

The tool chaining sentinel keys (`$tool`, `$path`, `$artifact`) are **never resolved** because the AI SDK validates tool args **before** calling `execute()`. The model's raw JSON `{ $tool: "..." }` fails schema validation and `execute()` (where `resolveArgs` lives) is never reached. The reference resolution system works correctly — it's just never invoked.

**What was changed:**
1. Use the AI SDK's `repairToolCall` callback to resolve references during the repair phase
2. **`imageInputSchema`**: Removed unnecessary `encoding` field — still a good cleanup but not the root cause

### DIFF
--- a/agents-api/src/domains/run/agents/generation/__tests__/ai-sdk-callbacks.test.ts
+++ b/agents-api/src/domains/run/agents/generation/__tests__/ai-sdk-callbacks.test.ts
@@ -1,8 +1,9 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { handlePrepareStepCompression } from '../ai-sdk-callbacks';
+import { agentSessionManager } from '../../../session/AgentSession';
+import { createRepairToolCallHandler, handlePrepareStepCompression } from '../ai-sdk-callbacks';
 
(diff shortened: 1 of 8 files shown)


tim-inkeep:
@mike-inkeep, how does the resolution fail validation because of sentinel keys yet is able to tool chain other items? Shouldnt all tool chainings be validated as well and have same error? If its image specific should we just update the image servers for what argument they receive?

mike-inkeep:
closing in favor of a different approach in an upcoming PR

PR #2756, opened 2026-03-18
fix: treat load_skill as internal tool to suppress false-positive Sentry errors

anubra266:
## Summary

- `load_skill` was being treated as a regular user-visible tool, so `wrapToolWithStreaming` emitted `tool_call`/`tool_result` graph events for it — but with no `relationshipId`, causing the graph animator in `use-animate-graph.ts` to fire a Sentry error on every skill load
- Fix: add `load_skill` to `isInternalToolForUi` alongside `transfer_to_*` and `delegate_to_*` — suppresses chat streaming events (input start/delta/output) and graph session events, matching other internal orchestration tools

Closes PRD-6271

### DIFF
--- a/agents-api/src/domains/run/agents/tools/default-tools.ts
+++ b/agents-api/src/domains/run/agents/tools/default-tools.ts
@@ -127,8 +127,7 @@ export async function getDefaultTools(
       'load_skill',
       createLoadSkillTool(ctx),
       streamRequestId,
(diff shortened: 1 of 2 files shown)


amikofalvy:
@claude Are there other such tools that can benefit from this pattern? How can we prevent this relationshipId issue from occurring again in the future?

PR #2793, opened 2026-03-20
reuse `useProjectsQuery` instead of `fetchProjectsAction` in `useEffect`

dimaMachina:


dimaMachina:
@claude --review and add a changeset

PR #2798, opened 2026-03-23
fix:removed tooltip provider at the root and used at the place wherever we are using the tooltip

### LINKED ISSUE #2786: [Performance] "New Scheduled Trigger" dialog causes severe FPS drops (5-7 FPS) due to background polling in table
## Problem
The "New Scheduled Trigger" dialog causes severe FPS drops (5-7 FPS) when opened, due to background polling firing every 3 seconds and re-rendering the dialog internals.

**Please check this Attached Video**
https://github.com/user-attachments/assets/23fda2fe-836f-478f-89d2-e1a8e176ad50

## Environment
- **Page:** `/{tenantId}/projects/{projectId}/triggers`
- **Machine:** Ubuntu, 32GB RAM, 16 CPU cores
- **Polling interval:** 3000ms
- **Tool used:** [[React Scan](https://github.com/aidenybai/react-scan)](https://github.com/aidenybai/react-scan) for detecting re-renders and FPS drops

## Steps to reproduce
1. Go to Triggers page with at least 1 scheduled trigger
2. Click "+ New Scheduled Trigger" to open the dialog
3. Wait 3 seconds for background poll to fire
4. Observe immediate FPS drop — no further interaction needed

## Observed via React Scan
- FPS drops to **5-7** every 3 seconds while dialog is open
- `Primitive.div` → **344ms** per re-render
- `Button` → **364ms** per re-render
- **7x FPS drops** recorded
- When dialog is closed, same polling fires with much less impact

## Expected
Opening the dialog should have no impact on background polling performance. Re-renders should stay under 16ms to maintain 60 FPS.

RajuGangitla:
Fixes #2786 

TooltipProvider was being mounted at the root level inside SidebarProvider, wrapping the entire app. Moved it to each individual Tooltip usage following the official shadcn/ui pattern. This avoids a single global provider re-rendering unnecessarily and aligns with the shadcn component guidelines.

### DIFF
--- a/agents-manage-ui/src/components/agent/node-library/node-item.tsx
+++ b/agents-manage-ui/src/components/agent/node-library/node-item.tsx
@@ -1,7 +1,7 @@
 import { GripVertical, type LucideIcon } from 'lucide-react';
 import type { ComponentProps, FC, ReactNode } from 'react';
 import { FlowButton } from '@/components/agent/flow-button';
-import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
 import { cn } from '@/lib/utils';
 
 export type NodeItem = {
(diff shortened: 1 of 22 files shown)


dimaMachina:
thanks for contribution, superseded by https://github.com/inkeep/agents/pull/2801

PR #2811, opened 2026-03-24
do not render tooltips for sidebar menu items

dimaMachina:
We previously had tooltips for sidebar items in `collapsed` mode

image

But in current design sidebar is expanded on hover and tooltip are never shown

This improve react-scan performance to eliminate 84 `Tooltip` and `TooltipTrigger` component render

## before
image

## after
image

### DIFF
--- a/agents-manage-ui/src/components/sidebar-nav/nav-item.tsx
+++ b/agents-manage-ui/src/components/sidebar-nav/nav-item.tsx
@@ -19,7 +19,8 @@ export function NavItem({ title, url, icon: Icon }: NavItemProps) {
       <SidebarMenuButton asChild isActive={pathname.startsWith(url)}>
         <Link href={url}>
           <Icon />
-          {title}
+          {/* Keep this span to prevent layout issues with long titles when sidebar collapsing */}
+          <span>{title}</span>
         </Link>
       </SidebarMenuButton>
     </SidebarMenuItem>


dimaMachina:
@claude add a changeset based on changes in this PR

RajuGangitla:
@dimaMachina This is amazing!

I’ve been trying a lot of ways since morning to reduce the tooltip renders in the sidebar, but I couldn’t find a proper solution.

I think the best approach is to remove them, since the sidebar expands on hover, so tooltips aren’t needed.

Sometimes the best optimization is simply removing things.

agents-manage-ui/src/components/sidebar-nav/nav-item.tsx
```diff
@@ -9,23 +9,17 @@ import type { IconComponentProps } from '@/components/ui/svg-icon';
...
   return (
     <SidebarMenuItem>
-      <SidebarMenuButton asChild tooltip={title} isActive={isActive}>
-        <Link
-          href={url}
-          rel={isExternal ? 'noopener noreferrer' : undefined}
-          target={isExternal ? '_blank' : undefined}
-        >
-          {Icon && <Icon />}
-          <span>{title}</span>
```
sarah-inkeep, line 28: I think we may need to keep this span here, I noticed that without it, long titles (like `API keys (Deprecated)` ) do a weird wrapping thing on hover (https://www.loom.com/share/23e08edfb53f4f15b33dff99ba6997d8)

agents-manage-ui/src/components/sidebar-nav/nav-item.tsx
```diff
@@ -9,23 +9,17 @@ import type { IconComponentProps } from '@/components/ui/svg-icon';
...
+      <SidebarMenuButton asChild isActive={pathname.startsWith(url)}>
+        <Link href={url}>
+          <Icon />
+          {title}
```
dimaMachina:
```suggestion
          {/* Keep this span to prevent layout issues with long titles when sidebar collapsing */}
          <span>{title}</span>
```

PR #2819, opened 2026-03-24
feat: add optional prompt field to app deployments

tim-inkeep:
## Summary
- Add an optional `prompt` field to app deployments that injects supplemental instructions into the agent's system prompt at runtime
- The app prompt appears as an `<app_context>` section after the agent context, allowing surface-specific behavioral tuning without duplicating agent configurations
- Includes schema migration, runtime threading, manage UI form updates, docs, and tests

## Changes
- **Schema**: New nullable `prompt` column on `apps` table (runtime DB migration)
- **Runtime**: App prompt threaded from auth middleware → execution context → system prompt builder → `<app_context>` template section
- **Observability**: `appPrompt` token count added to system prompt breakdown
- **UI**: Prompt textarea added to app create/edit forms
- **Docs**: App credentials docs updated with prompt field
- **Tests**: 7 unit tests (prompt assembly) + 6 integration tests (CRUD)

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All 2022 tests pass (including OpenAPI snapshot update)
- [x] New unit tests verify: prompt inclusion, omission (undefined/empty/whitespace), token tracking, section ordering
- [x] New integration tests verify: create with prompt, null default, update via PATCH, clear to null, GET/list responses


agents-api/src/__tests__/manage/routes/crud/apps.test.ts
```diff
@@ -552,4 +552,150 @@ describe('App CRUD Routes - Integration Tests', () => {
...
+      });
+      expect(res.status).toBe(201);
+
+      const body = await res.json();
+      expect(body.data.app.prompt).toBe(
+        'Be concise and link to documentation pages when possible.'
+      );
+    });
+
+    it('should create an app without a prompt (null by default)', async () => {
```
mike-inkeep: @claude this test is redundant with the test below it. If you agree, let's remove it

mike-inkeep:
@claude The test in executionHandler-app-prompt.test.ts should be deleted and replaced by tests added to runAuth-initiatedBy.test.ts.

  The deleted test verifies that executionContext.metadata.appPrompt gets spread into the A2AClient constructor headers — two lines of trivially simple conditional logic, exercised through ~15 mocks and a full ExecutionHandler.execute() run. The
  signal-to-noise ratio is poor, and the mock is incomplete (initializeStatusUpdates is missing, causing a silently swallowed error on every run).

  The new tests should cover the higher-value contract: that x-inkeep-app-prompt on an incoming request actually reaches executionContext.metadata.appPrompt through runAuth. This is where the feature can actually break silently — the header extraction and
   merge logic has two separate code paths and no prior test coverage. Getting that wrong means appPrompt never reaches ExecutionHandler at all, making the deleted test moot regardless.

  The other end of the chain — that appPrompt renders correctly into the system prompt as <app_context> — is already covered by PromptConfig.appContext.test.ts. Between that and the new runAuth tests, the meaningful contracts should be tested at the right
   layer boundaries.

PR #2821, opened 2026-03-24
feat(run): enable inline text document attachments via chat APIs

mike-inkeep:
## Summary

Extends the run chat APIs to accept inline base64 text document attachments (`text/plain`, `text/markdown`, `text/html`, `text/csv`, `text/x-log`). Accepted attachments are validated as UTF-8, stored in blob storage as URI-backed file parts, and replayed into model input as XML-tagged `<attached_file ...>` blocks — on the initial turn and through conversation history replay. Remote URLs remain disallowed for all non-image document types.

## Key decisions

**Inline-only, no remote URLs for text documents.** PDF URLs are still accepted, but text document attachments are restricted to inline `data:` payloads to avoid widening the remote-fetch attack surface. This was an explicit constraint from the start and is enforced at the schema validation layer.

**XML-tagged text blocks instead of provider-specific file parts.** Text documents are injected as `<attached_file filename="..." media_type="...">` XML blocks rather than as AI SDK file parts. This keeps injection behavior consistent across all model providers and ensures that the first-turn prompt and conversation-history replay produce identical prompts — no provider detection logic, no divergence.

**Blob-backed persistence only — no raw text in the DB.** Decoded attachment content is never written into conversation message rows. The blob URI and MIME type are what's persisted; the decoded text is fetched transiently at generation time and injected into model context. This matches the existing image/PDF pattern.

**256 KB decoded-byte cap per text attachment.** Chosen to prevent runaway token usage while still accommodating typical log, CSV, and Markdown files. Whether this is tight enough for strict prompt-budget protection is an open question flagged in the spec.

**`text/html` is raw source, not rendered text.** The API injects HTML source without sanitizing, converting, or rendering it. This means models see raw tags, not page text. This is a known tradeoff — noted in the PR description and flagged as a follow-up.

**MIME type is submitter-controlled.** The server does not infer text MIME type from filename extension. Clients must supply a correct `Content-Type` / media type. Server-side MIME fallback from extension is a deferred follow-up.

## Manual QA

## Future considerations

- Add server-side MIME fallback from filename extension for clients that upload text-like files with a missing or underspecified media type (e.g., `.log` uploaded as `application/octet-stream`).
- Consider whether additional text-like formats (JSON, XML) should follow the same inline-text attachment path.
- Revisit the 256 KB decoded-byte cap if prompt-budget pressure becomes a problem in practice.


agents-api/src/domains/run/agents/generation/conversation-history.ts
```diff
@@ -112,20 +120,55 @@ export async function buildConversationHistory(
   return { conversationHistory, contextBreakdown: updatedContextBreakdown };
 }
 
-export function buildInitialMessages(
+async function buildTextAttachmentPart(
```
tim-inkeep, line 128: was making this function async and converting the rest to be async just so we could build multiple text attachments asynchronously?

mike-inkeep: @tim-inkeep yes that's correct, buildInitialMessages() is now async so that we can fetch text attachments from blob storage and insert them there

PR #2827, opened 2026-03-25
feat: Authenticated chat sessions with asymmetric key verification

amikofalvy:
## Summary

Three related features for the Inkeep widget and agent framework:

1. **Authenticated chat sessions** — asymmetric key verification with customer-uploaded public keys
2. **Identity model enhancement** — verified token claims propagation, configurable authZ for global apps
3. **Playground unification** — playground uses the same app-credential auth path as customer widgets

### Authenticated Chat Sessions
- Zod schemas for `PublicKeyConfig`, `WebClientAuthConfig`
- Public key validation utility (algorithm allowlist, min key sizes, private key rejection)
- Key management API (`POST/GET/DELETE /apps/{appId}/auth-keys`)
- Asymmetric JWT verification in `tryAppCredentialAuth`
- Token requirements: `kid` (header), `sub`, `exp`, `iat` required. 24h max lifetime. 60s clock skew.
- Conversation identity: `userId` from verified `sub` claim
- Manage UI: auth keys section on app edit page
- Dual-mode support: apps with auth keys can still serve anonymous sessions (`allowAnonymous` flag, default true)

### Identity Model
- **Verified claims**: non-standard JWT claims extracted and propagated to `BaseExecutionContext.metadata.verifiedClaims` (runtime) and `conversations.metadata.verifiedClaims` (persisted)
- **Strict separation**: verified claims (signed) kept separate from unverified `userProperties` (client-side)
- **1KB limit** on verified claims
- **`validateScopeClaims`**: global apps can opt into SpiceDB validation of tid/pid claims
- **`buildConversationMetadata` helper**: shared function for conversation metadata construction

### Playground Unification
- Global app (`app_playground`) with `validateScopeClaims: true`
- Playground token endpoint signs JWTs in app-credential format (`kid`, `tid`/`pid`/`agentId` claims)
- Chat widget uses `x-inkeep-app-id` header (same path as customer widgets)
- PoW bypassed for authenticated sessions (JWT is the anti-abuse gate)

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| Asymmetric keys (not HS256) | No secrets on Inkeep. Unilateral key rotation. |
| Customer-uploaded public keys (not JWKS) | No hosting requirement. No network dependency. |
| Framework doesn't distinguish internal/external users | `sub` is always `endUserId`. `apiKeyId` identifies the source app. |
| Verified claims separate from userProperties | Different trust levels — signed token data vs. unverified client data |
| Global app SpiceDB validation is opt-in | Not all global apps need authZ checks — configurable via `validateScopeClaims` |
| Dual-mode (anon + auth) default on | Apps with auth keys still serve anonymous sessions unless `allowAnonymous: false` |

## Specs
- `specs/2026-03-23-authenticated-chat-sessions/SPEC.md`
- `specs/2026-03-26-identity-model-redesign/SPEC.md`

### DIFF
--- a/agents-api/__snapshots__/openapi.json
+++ b/agents-api/__snapshots__/openapi.json
@@ -10559,6 +10559,9 @@
       },
       "WebClientAuthConfig": {
         "properties": {
+          "allowAnonymous": {
+            "type": "boolean"
+          },
           "audience": {
             "type": "string"
           },
@@ -10569,6 +10572,9 @@
             },
             "maxItems": 5,
             "type": "array"
+          },
+          "validateScopeClaims": {
+            "type": "boolean"
           }
         },
         "type": "object"


agents-api/src/domains/manage/routes/appAuthKeys.ts
```diff
@@ -0,0 +1,274 @@
...
+  kid: z.string().min(1),
+});
+
+const AddPublicKeyBodySchema = z
+  .object({
+    kid: z.string().min(1).describe('Key identifier'),
+    publicKey: z.string().min(1).describe('PEM-encoded public key'),
+    algorithm: PublicKeyAlgorithmSchema.describe('Signing algorithm'),
+  })
+  .openapi('AddPublicKeyRequest');
```
amikofalvy: This is duplicate schema definition. We should already have defined this schema in validation/schemas.ts or in the same place that we define the schema for the db

agents-api/src/domains/manage/routes/appAuthKeys.ts
```diff
@@ -0,0 +1,274 @@
...
+          },
+        },
+      },
+      ...commonGetErrorResponses,
+    },
+  }),
+  async (c) => {
+    const { tenantId, projectId, appId } = c.req.valid('param');
+
+    const appRecord = await getAppByIdForProject(runDbClient)({
```
amikofalvy: One of the main benefits of drizzle orm (powers our DAL) is to write performant fetches. I think we should create a new DAL fetch function that only gets the keys needed instead of the entire appRecord. If we need any more of these functions in this route, please create them

agents-api/src/domains/run/routes/chat.ts
```diff
@@ -253,6 +257,9 @@ app.openapi(chatCompletionsRoute, async (c) => {
         activeSubAgentId: defaultSubAgentId,
         ref: executionContext.resolvedRef,
         userId: executionContext.metadata?.endUserId,
```
amikofalvy, line 267: Third party userIds should be independently tracked. I think we reserve the existing userId for inkeep userIds

agents-api/src/middleware/runAuth.ts
```diff
@@ -501,14 +592,20 @@ async function tryAppCredentialAuth(reqData: RequestData): Promise<AuthAttempt>
   }
 
   let endUserId: string | undefined;
-  let authMethod: 'app_credential_web_client' | 'app_credential_api';
+  let authMethod:
+    | 'app_credential_web_client'
+    | 'app_credential_api'
+    | 'app_credential_web_client_authenticated';
```
amikofalvy, line 603: duplicated types. Should reuse from elsewhere (zod infer?)

agents-api/src/domains/run/routes/chatDataStream.ts
```diff
@@ -252,6 +256,18 @@ app.openapi(chatDataStreamRoute, async (c) => {
           ref: executionContext.resolvedRef,
           agentId: agentId,
           userId: executionContext.metadata?.endUserId,
+          ...(body.userProperties || executionContext.metadata?.endUserId
```
amikofalvy: externalMetadata and internalMetadata.


Thinking through the Slack Work App:
- globalApp how does metadata get passed as external or internal?

PR #2853, opened 2026-03-26
Client-execution-mode

anubra266:


agents-manage-ui/src/components/agent/sidepane/section.tsx
```diff
@@ -25,15 +25,27 @@ export function SectionHeader({
...
+      {learnMoreHref && (
+        <a
+          href={learnMoreHref}
+          target="_blank"
+          rel="noopener noreferrer"
+          className="text-xs text-muted-foreground underline hover:text-foreground"
+        >
+          Learn more
+        </a>
+      )}
```
dimaMachina: let's avoid this prop in favour of composition pattern
https://github.com/inkeep/agents/blob/main/.agents/skills/vercel-composition-patterns/AGENTS.md

just pass link as `children`

PR #2890, opened 2026-03-29
remove `validateProjectId`, `validateTenantId`

dimaMachina:
these params already validated by zod


dimaMachina:
@claude add a changeset based on changes in this PR

PR #2946, opened 2026-04-01
Fix headersSchema case sensitivity for camelCase property names

nick-inkeep:
## Summary
- Normalize `headersSchema` property names and `required` entries to lowercase before AJV validation and key filtering
- Fixes 500 "Context validation failed" error when customers define camelCase properties (e.g., `mcpToken`) that don't match lowercased HTTP header keys (`mcptoken`)
- HTTP headers are case-insensitive per RFC 7230/9110 and normalized to lowercase by the Fetch API — the schema should match

### `validation.ts`
- Extract `normalizeSchemaKeysToLowercase()` — recursively lowercases JSON Schema property names and `required` entries
- Applied in `validateHttpRequestHeaders()` and `filterContextToSchemaKeys()` only (header-specific call sites)
- NOT applied in the shared `getCachedValidator()` — that function is also used by `ContextFetcher.validateResponseWithJsonSchema()` for API response schemas where property names are case-sensitive

### `contextValidation.headers.test.ts`
- 4 new test cases: camelCase properties, camelCase required entries, mixed case properties, camelCase filtering

### Other
- Changeset for `agents-api` (patch)
- Reverted unrelated auto-generated `evaluations.mdx` change

## Root cause
`contextValidationMiddleware` correctly lowercases incoming header keys (`key.toLowerCase()`), but the `headersSchema` property names were NOT lowercased before validation. AJV's property matching is case-sensitive, so `{ properties: { "mcpToken" } }` didn't match data key `"mcptoken"`. Additionally, `filterContextToSchemaKeys` iterates schema properties and checks `key in data` — also case-sensitive.

## Blast radius
- `normalizeSchemaKeysToLowercase` is scoped to header validation only
- `getCachedValidator()` (shared by ContextFetcher for API response validation) is NOT affected — response schemas with camelCase properties (e.g., `userId`) continue to work correctly
- Schema cache unaffected — cache key uses pre-normalization schema string

## Test plan
- [x] camelCase schema property validates against lowercase header
- [x] camelCase required entry enforces correctly
- [x] Mixed case properties all normalize
- [x] Filtering returns only declared (lowercased) properties
- [x] All 46 existing context validation tests pass (no regression)
- [x] Typecheck passes
- [x] Local review: APPROVE (low risk, 0 blocking issues)


nick-inkeep:
**Addressing review feedback:**

**Pullfrog (HIGH — shared getCachedValidator):** Already fixed in commit `91ad5b67b`. Normalization moved out of `getCachedValidator()` into `validateHttpRequestHeaders()` only. Delta review confirmed this is correct.

**Claude — nested object test (Minor):** Valid suggestion per spec test case #5. In practice, HTTP headers are flat key-value strings (not nested objects), so the recursive path is defensive code. The normalization function handles it correctly — adding a test for completeness would be a nice-to-have but not blocking.

**Claude — type guard edge case (Consider):** Matches existing `makeSchemaPermissive()` pattern (line 118) which uses the same `type === 'object'` guard. Consistency with the existing pattern is the right call — changing one without the other would create divergence.

**Claude + Pullfrog — evaluations.mdx (While You're Here):** This is auto-generated by Fumadocs during build. Will revert from this PR to keep the diff focused.

PR #3095, opened 2026-04-09
fix(agents-api): don't eagerly create text attachment artifacts

mike-inkeep:


tim-inkeep:
@pullfrog, add the changeset

